PR #10, opened 2025-04-04
Added deployment model doc and fixed broken link

YanisaHS:

PR #11, opened 2025-04-08
Add postgres extension info and minor change to upgrade guides

YanisaHS:


YanisaHS:
Note: The linkcheck will fail. This is fixed in open PR #10

Perfect5th (approved):

PR #12, opened 2025-04-09
Update juju installation doc

silverdrake11:


YanisaHS (approved):
Approved if you're happy with the changes I just pushed in. I kept the info, but reorganized a bit. I added a section back for the other bundles just to make that more clear (for anyone who's familiar with them). 

LMK if you have issues or questions about my changes, though :)

Also, I'm assuming the basic skeleton of the juju installation is all correct (Install, Deploy, Access), another engineer can review for that. I think it's okay for this doc to be quite high-level though, the HA one has more detail

PR #19, opened 2025-04-15
Reboot profile docs

silverdrake11:


srunde3 (approved):
LGTM, followed along with the docs and the inputs/responses were accurate.

docs/reference/api/rest-api-endpoints/beta-only.md
```diff
@@ -260,3 +260,220 @@ Example output:
...
+- `title`: The human readable title for this reboot profile.
+- `every`: The frequency at which the reboot should run. Only `"week"` is supported at this time.
+- `on_days`: A list of days (abbreviated: mo, tu, we, th, fr, sa, su) on which the reboot will run.
+- `at_hour`: The hour (0–23) at which the reboot will run. This time is interpreted in UTC.
+- `at_minute`: The minute (0–59) at which the reboot will run. This time is interpreted in UTC.
+- `all_computers`: Whether to apply this reboot profile to all computers.
+- `deliver_within`: Number of hours within which the task should be delivered.
+- `deliver_delay_window`: Randomize delivery within this number of minutes.
+- `access_group`: The access group name where the profile is stored.
+- `tags`: A list of computer tags to target instead of `all_computers`.
```
YanisaHS, line 348: Can you add the defaults if they're applicable to this endpoint too? (The previous one had defaults)

YanisaHS (changes requested):
Just one small comment to resolve, otherwise LGTM

docs/reference/api/rest-api-endpoints/beta-only.md
srunde3: This is a PATCH endpoint, so the default behavior is to not update a value unless provided. IMO that's implied by the PATCH

YanisaHS: Ahh, ok! Thanks for the info :smiley:

YanisaHS (approved):
LGTM :) Let's wait to merge though until beta is released

jansdhillon:
continued in #38

PR #23, opened 2025-04-23
Sanitize Instances Doc

HJK-X:
Adds the sanitize instance instructions.

### DIFF
--- a/docs/how-to-guides/web-portal/web-portal-24-04-or-later/index.md
+++ b/docs/how-to-guides/web-portal/web-portal-24-04-or-later/index.md
@@ -11,4 +11,3 @@ manage-access-groups
 use-annotations
 manage-livepatch-and-kernel-updates
 sanitize-instances
-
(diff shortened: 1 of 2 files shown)


YanisaHS (commented):
Great job! I've made some changes directly to your file. Please review the changes. LMK when you're ready for me to re-review, or if you're just good with those changes.

The changes are mostly just to make it a bit more clear, although I did add some brief explanation in the first sentence.

_Providing some context for my suggested changes below, wasn't expecting you to think these things when you created your first draft :)_

- Since the process for this guide is step-by-step, I preferred the numbered list approach rather than paragraph form. This is different from (for example) [this guide on using Livepatch in the web portal](https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/web-portal-24-04-or-later/manage-livepatch-and-kernel-updates/). That guide describes an entire tab, so explaining in that format works better for that type of how-to guide.
- I added a super brief description at the top. This is so users get a brief overview of what it is before they begin going through the guide
- I removed "from the Landscape web portal" from the H1. Normally, I don't add stuff like "from Landscape" or "from the Landscape web portal" unless it adds specific value to that guide. For example, that Livepatch one does say "from the Landscape web portal" because it's good to be clear in that case, so the user doesn't think they're going to do something with Livepatch directly

YanisaHS (approved):
Approved, but I'll merge it later

Issue #31, opened 2025-05-22
doc: Better to update the system before registering the client against Landscape Server


nobuto-m:
https://github.com/canonical/landscape-documentation/blob/8d5bd31e7cb2bc8fd408b6e6dfa25b7a61c7013b/docs/how-to-guides/landscape-installation-and-set-up/install-landscape-client.md?plain=1#L10-L14

There are some limitations that it takes time to synchronize installed/available package information when the system is not up to date. To maximize the smoothness of the initial Landscape experience, it would be good to make sure to run `apt upgrade` for example.

---
Document: getting-started-with-landscape.md

PR #32, opened 2025-05-22
Make sure the client system is up-to-date

nobuto-m:
There are some limitations in Landscape that it takes some time to synchronize the installed/available package information from the client to the server when the system is not up-to-date. To maximize the first user experience, make sure the system is up-to-date before registering clients to the server.

Closes: #31


nobuto-m:
I'm not 100% sure if `-y` is a good idea for `apt upgrade`. But I followed the style of the existing doc in the repo that non interactive commands are assumed in most of the areas.

YanisaHS:
Hey @nobuto-m! At this time, we don’t want to add the upgrade steps in the tutorial or runnable codeblocks as part of a general Landscape installation. This is because upgrading isn't actually *needed* to install or use Landscape (from a general audience’s perspective), upgrading can take awhile, and it’s not a good experience for users if they accidentally upgrade.

I’m going to close this PR but leave the issue open - this is good feedback that's still worth addressing in the documentation (e.g., as a note or something in some of the installation how-tos)

We still appreciate the contribution though! Thanks for your input :)

PR #33, opened 2025-05-22
Adding CI to notify staging of changes

YanisaHS:


.github/workflows/notify-staging.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  trigger-staging-sync:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install SSH key
+        uses: webfactory/ssh-agent@v0.9.0
+        with:
+          ssh-private-key: ${{ secrets.STAGING_DEPLOY_KEY }}
```
Perfect5th, line 16: What's the purpose of setting up an SSH agent if you use a github token for authentication in the next step? Where is SSH used?

Perfect5th (commented):
This approach seems fine to me, just curious what alternatives you considered.

Was it not possible to have this workflow push changes directly to the landscape-docs-staging repo, rather than having that repo pull from here?

.github/workflows/notify-staging.yaml
YanisaHS: This `STAGING_DEPLOY_KEY` is the private key that I added to the repo (and public was added to the staging repo). If I remove this part, then where does the permission to send dispatch events come from?

I'm definitely not sure if I'm setting it up correctly, though - the [`GITHUB_TOKEN`](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret) didn't quite make sense to me (GH suggested to [use it for authenticating](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28#authenticating-in-a-github-actions-workflow)), but the [repository dispatch event example](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event) uses a Bearer token instead of the `GITHUB_TOKEN`

I assumed the SSH part was also needed because if not, then couldn't anyone dispatch anything to our repo..?

YanisaHS: @Perfect5th Made those changes :)

Perfect5th (approved):
LGTM! Good luck testing it!

PR #51, opened 2025-07-02
docs: add grpc known issues

abehnia:


docs/reference/known-issues/index.md
```diff
@@ -0,0 +1,9 @@
+(reference-known-issues-index)=
+# Known Issues
```
YanisaHS, line 2:
```suggestion
# Known issues
```

docs/reference/known-issues/known-issues.md
```diff
@@ -0,0 +1,9 @@
+(reference-known-issues)=
+# Known Issues
```
YanisaHS, line 2:
```suggestion
# Known issues
```

docs/reference/known-issues/known-issues.md
```diff
@@ -0,0 +1,9 @@
...
+
+## gRPC Version
```
YanisaHS, line 4:
```suggestion
## gRPC version
```

docs/reference/known-issues/known-issues.md
```diff
@@ -0,0 +1,9 @@
...
+
```
YanisaHS, line 5: In this section, can you remove the extra line breaks? I know it renders the same on the docs site, but since our docs repo can have community contributions, it's important to keep this stuff consistent across the whole docs. We just use standard paragraph formatting

docs/reference/known-issues/index.md
YanisaHS, line 1: Couple general comments:

- This affects Landscape 23.10 and later for Jammy AND Noble, correct? If that's correct, then you can leave it as is - I don't think you need to specify the version. But if it's a more recent one (e.g., 24.10) or only affects Jammy, then you should add a note like "Since Landscape 24.10, ..." or "This impacts installations of Landscape 24.10 and later for Ubuntu Jammy installations"
- Is there anything the user should do to go around this that we could reasonably recommend? If yes, then it could be its own paragraph on the bottom. If there's nothing specific that makes sense for us to recommend, then leave as is. :)

YanisaHS (changes requested):

docs/reference/known-issues/known-issues.md
```diff
@@ -0,0 +1,9 @@
...
+Landscape installs its own version of `python3-protobuf` and `python3-grpcio` from its PPA, which are incompatible with the versions currently available
```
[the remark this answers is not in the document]
abehnia: We also install our packages on Noble so the stubs would be incompatible.

docs/reference/known-issues/index.md
abehnia: It started on 24.04 so I think it would be okay to leave it as it unless you think I should mention it. There's not much the user can do about this :disappointed:

wck0 (approved):
:ship: :it:

PR #52, opened 2025-07-03
Update configure-landscape-client.md

rajannpatel:
adding information about Repository Profiles to the documentation for Landscape Client


docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client.md
```diff
@@ -66,6 +66,23 @@ Once you’ve defined a registration key and enabled the auto-registration featu
 sudo landscape-config --account-name={LANDSCAPE_ACCOUNT_NAME} --computer-title={COMPUTER_TITLE} --registration-key={KEY}
 ```
 
+(howto-heading-repository-configuration-enforcement)=
+## Repository configuration enforcement
+
+By default, when a Repository Profile is applied to a machine, it enforces the `/etc/sources.list` configuration on the Ubuntu instance to match exactly what is specified in Landscape Server. The Repository Profile configurations are applied once, and are not validated on an ongoing basis. Instead of replacing repository configurations in /etc/sources.list, Landscape Client can append the additional repository configurations defined the Repository Profile, if the `manage_sources_list_d` configuration in `/etc/landscape/client.conf` is set to false.
```
YanisaHS, line 72:
```suggestion
By default, when a repository profile is applied to a machine, Landscape enforces the exact `/etc/apt/sources.list` configuration as defined in the profile. This configuration is applied once when the profile is applied, and is not revalidated or enforced on an ongoing basis. 

If you want Landscape to append the repository configurations instead of replacing the existing ones in `/etc/apt/sources.list`, you can configure Landscape to use `/etc/apt/sources.list.d` instead by setting the `manage_sources_list_d` configuration in `/etc/landscape/client.conf` to false.
```

1. I've changed the paths to have `apt` - is this correct? (I found this in the [repository management API guide](https://documentation.ubuntu.com/landscape/how-to-guides/repository-mirrors/manage-repositories-with-the-api/#add-pockets-to-the-repository-profile))
2. Is this meaning still accurate? I've mostly changed the second part because I didn't think it was clear what was actually happening when the configuration is set to false. But I'm not totally sure that I've understood what happens here correctly

docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client.md
```diff
@@ -66,6 +66,23 @@ Once you’ve defined a registration key and enabled the auto-registration featu
 sudo landscape-config --account-name={LANDSCAPE_ACCOUNT_NAME} --computer-title={COMPUTER_TITLE} --registration-key={KEY}
 ```
 
+(howto-heading-repository-configuration-enforcement)=
+## Repository configuration enforcement
```
YanisaHS, line 70:
```suggestion
## Control repository configurations
```
Thoughts? I'd rather it starts with a verb, but I'm open to other rewordings

YanisaHS (changes requested):
A few suggested changes. I'm going to ask an engineer to technical review as well

docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client.md
HJK-X: when the `manage_sources_list_d` configuration is set to false, Landscape Client will not modify the configuration in /etc/apt/sources.list.d

maybe:
```
If you do not want Landscape to delete the repository configurations in `/etc/apt/sources.list.d`, you can configure Landscape Client to only modify `/etc/apt/sources.list` instead by setting the `manage_sources_list_d` configuration in `/etc/landscape/client.conf` to false.
```

YanisaHS: Thanks Justin! So @rajannpatel the bottom part could be what he suggested instead

YanisaHS: @rajannpatel The last paragraph should be Justin's suggestion though - not the one I originally posted

YanisaHS (changes requested):
There's one comment I've left in a previous thread

YanisaHS (changes requested):
Mostly LGTM, except you need to add "revalidated" to the [custom_wordlist.txt](https://github.com/canonical/landscape-documentation/blob/main/docs/.custom_wordlist.txt) to pass the spelling check

HJK-X (commented):
LGTM

YanisaHS (approved):
LGTM! I'll merge this for you

PR #58, opened 2025-07-15
Add Google Analytics and cookie note

YanisaHS:


YanisaHS:
There was an error in this PR, fixed in #59

PR #64, opened 2025-08-11
Add doc comparing Landscape and Ansible

YanisaHS:


kmetscher (commented):
This is very clear and communicates the differences between the two in a way that people familiar with Ansible are probably going to understand immediately, especially by highlighting the distinction between the agent and agentless approaches and the PIT vs ongoing statefulness each system uses. LGTM, I have nothing to add and think this is suitable for publishing if Mark +1s.

mdscunningham:
I agree with Kyle by and large. I think this is very clear and concise and communicates the major points in a way that is easily accessible and understood. 

I would maybe quibble with the "real-time data and control" in the comparison table. I would rather this be more clear about the asynchronous nature of client/server communication, but I'm not sure exactly how I would word that. 

I'll add another comment if I think of something especially elegant or clever to put there instead.

PR #65, opened 2025-08-14
Refresh homepage

YanisaHS:
Note that I still have to follow the similar structure as the original for consistency across Canonical's docs, but I've updated the doc to make more clear what Landscape does.

The structure I have to follow is essentially:
Paragraph 1: What is Landscape? 
Paragraph 2: How does Landscape work? 
Paragraph 3: What need does Landscape meet? 
Paragraph 4: Who is Landscape for?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Landscape Documentation
 
-Landscape is a systems administration tool for managing and monitoring Ubuntu deployments, ranging from a few machines to large environments. It lets you control configuration, updates, and security from a central location instead of handling each machine separately. You can use Landscape for all of your Ubuntu systems, such as desktops, servers, cloud instances, and IoT devices.
+Landscape is a systems administration tool for managing and monitoring Ubuntu deployments, ranging from a few machines to large environments. It lets you control configuration, updates, and security from a central location. You can use Landscape for all of your Ubuntu systems, such as desktops, servers, cloud instances, and IoT devices.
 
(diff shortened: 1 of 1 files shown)


astok (changes requested):
typo: "on-premises" should be "on-premise", I believe.

README.md
```diff
@@ -1,12 +1,12 @@
 # Landscape Documentation
 
-Landscape is a systems management tool designed to help you manage and monitor your Ubuntu systems from a unified platform.
+Landscape is a systems management tool for managing and monitoring Ubuntu deployments, ranging from a few machines to large environments. It lets you control configuration, updates, and security from a central location instead of handling each machine separately.
```
yurii-vasyliev, line 3: As discussed earlier, I'd rather start the sentence with something like "Landscape is a systems administrative tool for managing and monitoring..." or "Landscape is a systems administrative tool that helps you manage and monitor...".
I'd also mention the different types of instances we support, like "standalone" Ubuntu instances, WSL, etc.

rajannpatel, line 3: It lets you push security and bugfix updates, as well as configure, harden, audit, and inventory software and users.

YanisaHS: I've removed the part you suggested about handling machines separately, but I kept the earlier parts only because your suggested wording is very similar to what's in the second paragraph, so it didn't read well when all put together. The content of what you've suggested is still in the doc though

PR #66, opened 2025-08-15
add package-reporting explanation doc

Perfect5th:
Migration of https://github.com/canonical/landscape-internal-docs/pull/175

has been approved there.


YanisaHS (changes requested):
You need to add the index file to `features` - the H1 will show up on the sidebar

And also add a ref target before your H1 in the package-reporting doc

PR #85, opened 2025-09-24
Update contributor guide

YanisaHS:
I've added information on how to actually contribute (pull the docs down locally, test, etc..). In the future, I believe the Documentation team will create a standard contribution guide, so I didn't want to spend _too_ much time on it. But this information still needs to be there because we don't currently have any guidance on this


astok (approved):
LGTM

PR #90, opened 2025-10-08
Ignore Hashicorp vault link due to rate limiting causing test to fail

YanisaHS:


YanisaHS:
I've added the Hashicorp link to our linkcheck ignore because it's consistently failing the check, although the link works if you manually go to it. This isn't unexpected because it's becoming increasingly common for sites to set up stricter controls to protect their actual users due to abusive chatbot behavior

(The PR is a little misleading because the linkcheck still failed, but due to wiki.debian.org being down, not the ignored Hashicorp link. I don't want to add address that debian link unless it becomes a consistent issue, such as this Hashicorp one)

PR #98, opened 2025-10-23
feat: add tokens API reference and authn/authz explanation

srunde3:


docs/explanation/security/authentication-and-authorization.md
```diff
@@ -0,0 +1,66 @@
...
+## Authentication model
+
+Prior to 24.04, Landscape used sessions for UI authentication and an access key/secret pair for API authentication. Beginning in 24.04, Landscape introduced JWTs for API and UI authentication.
+
+|         | 24.04                         | 24.10                         | 25.04                         | 25.10                         |
+|---------|-------------------------------|-------------------------------|-------------------------------|-------------------------------|
+| Old UI  | Session only                  | Session <br> or JWT           | Session <br> or JWT           | Session <br> or JWT           |
+| New UI  | JWT only                      | JWT only                      | JWT only                      | JWT only                      |
+| v1 API  | Access key/secret <br> or JWT | Access key/secret <br> or JWT | Access key/secret <br> or JWT | Access key/secret <br> or JWT |
+| v2 API  | JWT only                      | JWT only                      | JWT only                      | JWT only                      |
```
YanisaHS: A few things:
- Change 24.04 to "24.04 LTS"
- "Old UI" should be "Classic UI" because this is what we call it in the docs, and also it's the main UI for a lot of people still. Normally I wouldn't like "new" UI, but I think it's fine for here since it's not the default UI yet.
- v1 should be Legacy API and v2 should be REST API

docs/explanation/security/authentication-and-authorization.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Sessions
+
+Landscape sets a session with the browser when a user logs into the old UI. It is cleared when the user logs out of either the old or new UI. A session is only a valid access credential on the old UI. It is not accepted in the new UI or in any APIs.
+
+### JWTs
+
+Landscape issues a JWT and stores it as a cookie when a user logs into the new UI. The cookie is cleared when the user logs out of either the old or new UI. JWTs are valid access credentials for the old and new UIs, and both APIs.
```
YanisaHS:
```suggestion
Landscape issues a JWT and stores it as a cookie when a user logs into the new UI. The cookie is cleared when the user logs out of either the classic or new UI. JWTs are valid access credentials for the classic and new UIs, and both APIs.
```

docs/explanation/security/authentication-and-authorization.md
```diff
@@ -0,0 +1,66 @@
...
+
+See {ref}`explanation-cryptographic-technology` for details on JWT encryption. JWTs are used for identity only; they do not contain any authorization information.
+
+JWTs expire after 24 hours. JWTs can be invalidated for a user, an account, or the entire system. See {ref}`reference-rest-api-tokens`.
+
+### Access key/secret pairs
+
+Access key/secret pairs are issued when a user creates API credentials in the old UI. Access key/secret pairs may only be used to access the v1 API. However, they can be exchanged for a JWT that is valid in the v2 API. See {ref}`reference-rest-api-login`.
```
YanisaHS:
```suggestion
Access key/secret pairs are issued when a user creates API credentials in the classic UI. Access key/secret pairs may only be used to access the legacy API. However, they can be exchanged for a JWT that is valid in the REST API. See {ref}`reference-rest-api-login`.
```

docs/explanation/security/authentication-and-authorization.md
```diff
@@ -0,0 +1,66 @@
...
+
+Access key/secret pairs may be regenerated at any time, which will invalidate the old credentials. See {ref}`reference-rest-api-credentials`.
+
+## Authorization model
+
+Landscape uses role-based access control (RBAC). Users receive roles, and these roles are mapped to a set of permissions. Permissions determine the actions that a user is authorized to perform in Landscape.
```
YanisaHS:
```suggestion
Landscape uses role-based access control (RBAC). Users are assigned roles, and these roles are mapped to a set of permissions. Permissions determine the actions that a user is authorized to perform in Landscape.
```
This sounded more precise to me, but feel free to disagree.

docs/explanation/security/authentication-and-authorization.md
```diff
@@ -0,0 +1,66 @@
...
+
+Landscape administrators can create custom roles and map them to which permissions they choose. See {ref}`how-to-web-portal-manage-admins-and-roles`.
+
+## Email uniqueness
+
+Landscape requires unique emails for all users. A user created through local email/password is considered a distinct user from one created through and external authentication method, even if the email is the same. Therefore, it is not possible to create an externally-authenticated user with the same email as a local user. Similarly, it is not possible to create two users with the same email from two different external authentication methods.
```
YanisaHS:
```suggestion
Landscape requires unique emails for all users. A user created through local email/password is considered a distinct user from one created through an external authentication method, even if the email is the same. Therefore, it is not possible to create an externally-authenticated user with the same email as a local user. Similarly, it is not possible to create two users with the same email from two different external authentication methods.
```

YanisaHS (changes requested):

PR #103, opened 2025-10-29
25.10 release docs 

YanisaHS:
This PR moves the items from our internal staging repo into the public main repo.

**For reviewers:**

I don't expect you to look through the items in depth. There were no merge conflicts, so I haven't changed any of the user docs. At a minimum, you should scan through the files changed. (I have also done this and checked for the docs in the RTD build)

The only thing I changed is in the last two commits - I've changed some parts of the workflows because this is the first time we did this process, and they differed between the two repos. I've now made them match, but with conditions about which repo they run in. (also, if we find an issue with the workflows later, that can be solved outside of this PR)

### DIFF
--- a/.github/workflows/notify-staging.yaml
+++ b/.github/workflows/notify-staging.yaml
@@ -1,0 +1,20 @@
+# .github/workflows/notify-staging.yml
+name: Notify Staging Repo
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  trigger-staging-sync:
+    if: ${{ github.repository == 'canonical/landscape-documentation' }}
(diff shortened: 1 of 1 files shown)


YanisaHS:
I've looked at the PR pretty thoroughly and tested it already. I'll get this reviewed but wait to publish until everything else is in place

Perfect5th (approved):
Went through the file list and scanned the diffs. Didn't seen anything that seemed off.

Let me know if there's anything more specific you'd like checked.

YanisaHS:
Great, thanks! I'll wait to merge this in until other parts of the release are ready

PR #108, opened 2025-11-04
Update 25.10-release-notes.md

rajannpatel:
additional details for WSL and Provisioning


YanisaHS (changes requested):
@rajannpatel Regarding the Ubuntu Installer part - we have [a note in the relevant doc](https://documentation.ubuntu.com/landscape/how-to-guides/ubuntu-installer/provision-a-workstation/) that it's an experimental feature in the Ubuntu Installer. AFAIK, this isn't documented in Ubuntu Installer docs, so I'm hesitant to include this information here.

Can you either just remove it, or can you remove it AND add a line at the end of the paragraph like "Note that this is an experimental feature in the Ubuntu Installer." (or similar)

YanisaHS (approved):

PR #122, opened 2025-11-26
Add note on version

YanisaHS:
**For reviewer:**

Please check that the version is correct. I think it is based on the [24.10 release notes](https://documentation.ubuntu.com/landscape/reference/release-notes/24-10-release-notes/)


YanisaHS:
Ignore the spellcheck - it'll break but I have a change to fix those issues in another PR

PR #125, opened 2025-12-03
consolidated profiles

YanisaHS:
**For reviewers:**

- I've tested that the redirects work
- There are no changes to the text in the individual profiles (files I deleted). The only text I added is the intro where I defined what a profile is
- Refreshing the language in the individual profile descriptions will come in later work this cycle - this PR is only to consolidate those into one page and add redirects. (I.e., don't review those explanations - it'll get worked on later)


JacobisEpic (approved):
LGTM! :ship: :shipit:

PR #134, opened 2025-12-10
feat: add reference for wsl profiles

wck0:


docs/reference/terms/profiles.md
```diff
@@ -25,3 +25,15 @@ A **removal profile** defines a maximum number of days that a computer can go wi
...
+
+(reference-terms-wsl-profile)=
+## WSL profile
+
+A **WSL profile** defines what Ubuntu WSL instances should be installed on Windows instances associated with a specific access group. You can associate zero or more Windows instances with each WSL profile via tags to install Ubuntu WSL instances on those Windows instances. Each WSL profile defines a single Ubuntu WSL instance that should be present on associated Windows instances. The profile can specify which Ubuntu distribution to install, either from the Windows store or from a custom image, optionally configured with with cloud init. You can also configure the WSL profile to be **strict** which requires the Windows instance to only have Ubuntu WSL instance that are created through Landscape.
+
+WSL profiles are evaluated periodically for **compliance**. A Windows instance is deemed to be **compliant** if both of the following are true:
```
YanisaHS:
```suggestion
WSL profiles are evaluated periodically for compliance. A Windows instance is considered **compliant** if the following are true:
```
- I think bolding the second is sufficient. I try to limit the bolding in general because it can get to be too much sometimes
- Simplified language
- Removed "both" - this makes it more maintainable w/o sacrificing the language anyway. (Like if another option got added later, it won't look stale if that person misses changing "both")

docs/how-to-guides/wsl-integration/use-wsl-profiles.md
```diff
@@ -38,11 +26,14 @@ Then complete the relevant fields:
...
+
     ```powershell
     $distributionListUrl = "https://raw.githubusercontent.com/wiki/canonical/ubuntu-pro-for-wsl/beta/manifest.json"
     Set-ItemProperty -Path "HKLM:SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss" -Name DistributionListUrl -Value "$distributionListUrl" -Type String -Force
     ```
+
 - **cloud-init** (optional): The contents of the cloud-init to be supplied to the WSL instance. This can be uploaded as a file or inputted as text. If you don't include a cloud-init file, the rootfs image will be applied without any additional configuration.
+- **Compliance settings** : Whether or not the associated Windows instances should only have WSL instances that were created through Landscape.
```
YanisaHS: Ahh - I see this here. Going to go edit my comments on below (in case you read this first, I was asking about the concept of "strictness", which is new)

PR #137, opened 2025-12-10
refactor: clean up reference docs

wck0:


wck0:
I think I fixed them all.

> I've had it happen before and wasn't sure what part of the code within the block was invalid. If needed, having blocks "miscategorized" (e.g., as `bash` instead of `json`) isn't _that_ big of a deal, although having the correct one is ideal. But if you are unsure how to fix the block, just leave those ones in their original form

FWIW, some of the errors come from code blocks like this one: https://github.com/canonical/landscape-documentation/blob/main/docs/reference/api/legacy-api-endpoints/alerts.md?plain=1#L78-L104

There are two code blocks there; the first looks like Python code and the second is JSON. I'm not sure we need both, since the API response will be JSON. My guess is they were copied over from old docs that had just copied the docstring of the API method.

I think it makes sense to delete the first code block in that example as it's both redundant and confusing to have both. If you agree, I can make a follow up PR to scrub them.

PR #139, opened 2025-12-11
feat: script profiles definition

JacobisEpic:
## Context
https://warthogs.atlassian.net/browse/LNDENG-3263


docs/reference/terms/profiles.md
YanisaHS: This isn't just a script profiles thing, but I noticed all the others call it the Profiles "tab", which is relevant to our old UI. The old UI had these as tabs, but the new UI has a Profiles page in the sidebar.

I try to keep things consistent because it can be confusing to call multiple things "tabs" in the UI - the horizontal ones (like where script profiles is) are tabs, but the ones in the sidebar, we call those pages (or just specify the name, like in my example)

Can you change the language in the other profiles on this page to remove the "tab" and "organization's home page" part?

So:

**Before**: "from the **Profiles** tab in your organization's home page."

**Change this to**: "from the **Profiles** page."

PR #141, opened 2025-12-15
feat: Security profiles definition

JacobisEpic:
## Context
https://warthogs.atlassian.net/browse/LNDENG-3264


YanisaHS:
You'll probably get a legitimate linkcheck fail for the Enterprise Store docs btw - just ignore it. They're in the middle of testing stuff and they're down now

PR #143, opened 2025-12-16
new guide to create a saas account with our signup page and register the first client

YanisaHS:


YanisaHS:
Note: The Enterprise store docs may fail the linkcheck - they're testing something with it right now

HJK-X (approved):
LGTM :tada:

PR #160, opened 2026-02-01
fix: refresh tags explanation

jmucc:
https://warthogs.atlassian.net/browse/LNDENG-3258

- Add example
- Add instructions for new UI and API


YanisaHS (changes requested):
Couple small things but it mostly looks good!

PR #189, opened 2026-04-08
Refresh web-based repo mirroring doc

YanisaHS:
Note that this is still to represent the experience in 24.04. The 26.04 repo mirroring documentation will come later


docs/how-to-guides/repository-mirrors/manage-repositories-in-the-web-portal.md
```diff
@@ -165,6 +149,8 @@ Note that you may want to create multiple repository profiles for different grou
 
 ## Create and manage pull pockets
 
+A pull pocket is a user-defined pocket that acts as a staging area for selected packages and updates from another pocket, so you can control what gets distributed to different groups of machines.
```
david-mclain, line 152: All of the docs around pull pocket's I'm not very familiar with as I have not touched reprepro at all (outside of deleting code) so I cannot confidently confirm if these details are correct

david-mclain (approved):
LGTM, just a comment left around the fact I am unfamiliar with some of the concepts that are went over in this doc. If we would like to get a second opinion to ensure everything is correct then that'd be best but outside of that everything seems good.

docs/how-to-guides/repository-mirrors/manage-repositories-in-the-web-portal.md
YanisaHS: Thanks for calling it out! The pull pocket part is copied from another doc in our public docs, so I can confirm that's correct :)

PR #197, opened 2026-04-22
refresh access groups explanation

jmucc:
- Added some details
- Added instructions for the new web portal
- Added instructions for the legacy API
- Used more recent terminology ("instances" instead of "computers")


YanisaHS (approved):
LGTM! Appreciate your meticulous work with cross-linking to the other pages :)

Before you merge though - that linkcheck failure is an actual broken link, can you remove [this line](https://github.com/canonical/landscape-documentation/blob/main/docs/how-to-guides/repository-mirrors/manage-repositories-with-the-api.md?plain=1#L82) from that repo management API guide? I don't see an issue with removing the whole line since it's a link to a manpage, which is intuitive (and often done from the CLI anyway). We don't need to provide that guidance

PR #199, opened 2026-04-23
Update Self-hosted Landscape explanation and increase visibility of PPA docs

YanisaHS:
The two pages are similar, but the reasoning for keeping that information in the "About self-hosted Landscape" page is that that's a popular entry point for Field Engineering. And the second issue I was solving is that the PPA information isn't in a logical place in the docs, although I didn't want to remove it from that page either, so I chose to duplicate it in two places.


jansdhillon (approved):
LGTM

PR #207, opened 2026-04-30
update repository mirroring section for 26.04/debarchive

Perfect5th:


YanisaHS:
@Perfect5th (just saw this PR) but since we don't have actual versioned docs yet - instead of removing older content - can you make this as a completely separate guide? It can just go right below the other one and copy the same text as needed, named something with the versions (e.g., "Repository mirroring (Landscape 26.04 LTS+)"), with a note at the top of each linking the two ("If you're using Landscape 26.04 LTS+, see [link] instead"; "If you're using Landscape 25.10 or earlier, see [link] instead")

It's clunky for our current users, but we'll version hopefully soon and then just have one guide per version of the docs.

This would be the case for any major docs like this btw - so we don't remove the docs for older versions of Landscape.

(also if this is still a draft can you mark it that way?)

Perfect5th:
I've rearranged things such that there are now two pages that reference each other, one for >=26.04 and the original one for <=25.10.

PR #210, opened 2026-05-01
feat: add outbox configuration reference

srunde3:
This documents how to configure the `landscape-outbox` snap. This snap is required for `landscape-server` 26.04.

The snap has default configuration that allows it install seamlessly alongside an existing Landscape installation; I don't expect the "average" user to need to use this reference.

### DIFF
--- a/docs/explanation/landscape/landscape-server-architecture.md
+++ b/docs/explanation/landscape/landscape-server-architecture.md
@@ -11,7 +11,7 @@ Landscape Server is the server-side component of the Landscape ecosystem. It is 
 
 ![Landscape Service Diagram](/assets/images/landscape-services.jpg "Landscape Services")
 
-A Landscape Server deployment has six required services:
+A Landscape Server deployment has seven required services:
 
 * [API](#api) - serves REST API and Legacy API requests
 * [Appserver](#appserver) - serves Legacy UI and static files for the new UI
@@ -19,6 +19,7 @@ A Landscape Server deployment has six required services:
 * [Job handler](#job-handler) - runs background jobs such as repository mirror syncs
 * [Message system](#message-system) - exchanges messages with Landscape Clients
 * [Pingserver](#pingserver) - records Landscape Client heartbeat pings
+* [Outbox](#outbox) - ensures reliable, eventually-consistent delivery of events across databases and the message broker
 
 There are also optional services. Without these, Landscape Server is usable, but certain features will not be available:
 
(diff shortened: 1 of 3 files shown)


srunde3:
@wck0 don't necessarily have to review this, but having a configuration reference came out of the last outbox PR. Pinging in case you're interested.

YanisaHS:
@srunde3 if this is for the 26.04 release, you should re-target it to the `release/26.04` branch

**Edit:** Just noticed the description - I'll fix the branch for you!

YanisaHS (approved):
It LGTM - some general comments though (up to you if you want to change it now/in this doc):

The doc isn't clear about when you'd use the outbox snap environment variables vs when to use what's in service.conf or how the user would think about this, although arguably that information doesn't belong in the config reference anyway - just sharing my thoughts on that as a topic

IMO though, it'd still be helpful to add something concise that ties the concepts together, like "You can either set environment variables directly or provide them in a `service.conf` file. When a `service.conf` file is used, its values are mapped to the corresponding environment variables used by the outbox service." (assuming that's true) - to give the user a more direct nudge about. If you add it, I'd swap or merge the contents with "Both the outbox and cleanup services can read a Landscape service.conf file and translate its values into the environment variables that the service expects."

Up to you though! Feel free to re-request a review if you want me to look at anything else here

srunde3:
@YanisaHS maybe I should put together a separate how-to/explanation on the outbox then? IMO the important concepts to get across in another document(s) are:

- The outbox is necessary for correct operation of Landscape 26.04
- The outbox needs to connect to the same databases and brokers as `landscape-server`
- The outbox integrates with an existing Landscape server installation by reading and adapting its service.conf
- The outbox is deployed as a snap and may be configured through `snap set landscape-outbox ...`

Does this feel how-to-ish or explanation-ish?

YanisaHS:
@srunde3 It's explanation - super distilling your points down to you want to explain (1) what it is, (2) why it's there, (3) how it connects with existing stuff in Landscape

The `snap set` part could be described in a general sense in an explanation, doing something specific (like setting a specific config) would be in a how-to (which already is in our setup guides, not sure if anything else would be needed now) - providing that context because presumably that was the area that's unclear if it's how-to or explanation

Presumably your explanation would go here though? https://documentation.ubuntu.com/landscape/explanation/landscape/landscape-server-architecture/

srunde3:
@YanisaHS I added a brief explanation and made the recommendation to use the service.conf integration a bit stronger. LMK what you think.

YanisaHS (approved):
Yes I like this better 🥇

YanisaHS:
@srunde3 WAIT YOU HAVE A SPELLING ERROR

PR #221, opened 2026-05-19
TSG for common Landscape client issues

YanisaHS:


docs/how-to-guides/troubleshooting/common-landscape-client-issues.md
```diff
@@ -0,0 +1,562 @@
...
+
+If the service shows `Start condition unmet`, see {ref}`heading-client-start-condition-unmet`.
+
+### Check the broker log
+
+```bash
+sudo tail -n 50 /var/log/landscape/broker.log
+```
+
+The broker log is usually the best first log to check for client registration, certificate, hostname resolution, and client-server communication issues.
```
jmucc: nit: I `"best first"` sounds confusing to me. I think keeping one adjective or the other would increase clarity.

docs/how-to-guides/troubleshooting/common-landscape-client-issues.md
```diff
@@ -0,0 +1,562 @@
...
+- If the log contains an error referencing `prostatus.json`, see {ref}`heading-client-pro-status`.
+- If the client is registered but doesn't appear online or doesn't pick up activities, see {ref}`heading-client-debug-logging`.
+- If package status is incorrect or stale, see {ref}`heading-client-package-reporting`.
+
+### Check hostname resolution
+
+Check that the Landscape server hostname resolves correctly:
+
+```bash
+getent hosts <LANDSCAPE-SERVER-NAME>
```
jmucc, line 49: What do you think about switching this to `$LANDSCAPE_FQDN`. I think it's more commonly used in our docs for this concept (but that doesn't mean it's better necessarily)

docs/how-to-guides/troubleshooting/common-landscape-client-issues.md
```diff
@@ -0,0 +1,562 @@
...
+sudo landscape-config
+```
+
+If `/etc/landscape/client.conf` already contains the required account and server configuration, you can reuse the existing configuration:
+
+```bash
+sudo landscape-config --silent
+```
+
+Only use `--silent` when the required configuration values are already present in `/etc/landscape/client.conf` or provided as command-line options.
```
jmucc, line 399: I was thinking about this earlier in the doc when suggesting users to run `sudo landscape-config`. I think it applies in most cases but I am not sure it's worth repeating all over the doc.

jmucc (commented):
Couple small things. I think you can take or leave them.

docs/how-to-guides/troubleshooting/common-landscape-client-issues.md
YanisaHS: The reason to repeat it is that users won't be reading the whole guide in order like you and I did :) they'll likely just be in whatever section is relevant to their issue

docs/how-to-guides/troubleshooting/common-landscape-client-issues.md
YanisaHS: I think it's a good suggestion for another type of doc, but the reason I wouldn't add it here is it requires the variable is set (whereas `<>` denote a placeholder variable), which makes sense for a how-to guide where someone goes through the whole thing and there can be common instructions at the top, but this one is more likely for a user to go directly into a specific section, so it'd need to be set again multiple times and isn't that much of a shortcut.

also AFAIK it doesn't _need_ to be the FQDN but can be the hostname (but correct me if I'm wrong on that!)

jmucc:
LGTM

jmucc (approved):

PR #225, opened 2026-05-21
update repository mirroring section for 26.04/debarchive

Perfect5th:
Re-targeting of https://github.com/canonical/landscape-documentation/pull/207, which seems to have been dropped somehow.


docs/explanation/features/repository-mirroring-2604.md
YanisaHS, line 1: (going to go through the guide much closer and more suggestions but)
At first pass, I have a couple main points:

1. "debarchive" in general seems overexposed for this explanation. The previous one doesn't mention reprepro at all. I think it's fine if there's some callout (I'll put that as a suggestion somewhere), but debarchive doesn't really _need_ to be exposed here. It should come somewhere in the docs (like the server architecture guide)
2. The migration knowledge / mapping of concepts should be somewhere else too - probably in the repo management migration guide that's coming later. It's not really relevant to Landscape 26.04 (except in the case of people migrating from previous versions)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+```{note}
+This document applies to **Landscape 26.04 and later**. If you're using Landscape 25.10 or earlier, see {ref}`repository mirroring <explanation-repo-mirroring>` instead.
+```
+
+Landscape uses repository mirroring to internally distribute software in Debian packages over your local network and manage updates. This feature allows you to establish custom repositories from your local mirror, which provides an additional layer of control over the software versions available to your client machines. This can also reduce bandwidth usage by distributing updates to clients using your local network.
+
+You can mirror publicly accessible Ubuntu repositories, third-party repositories, or create local repositories for distributing internal software.
+
+Repository mirroring in Landscape is backed by the **landscape-debarchive** service, which must be installed alongside `landscape-server`. This service manages mirrors, local repositories, publications, and storage backends through an API that Landscape Server and the Landscape Server UI consume.
```
YanisaHS:
```suggestion
```
I'd recommend cutting this here - I'll suggest a new one below, at the bottom of the intro

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Snaps can't be mirrored with Landscape. If you want to mirror snaps, use [Enterprise Store](https://documentation.ubuntu.com/enterprise-store/main/).
+
```
YanisaHS, line 19:
```suggestion
In Landscape 26.04 LTS and later, repository mirroring uses a publication-based model. You create or sync repository content, publish that content to a target, and configure client machines to use the published repository as an APT source.

The service that provides this functionality is `landscape-debarchive`. You might see this name in installation, configuration, service status, logs, and troubleshooting materials.
```
Something like this? (reword as necessary) - and then `debarchive` doesn't get mentioned anywhere else in this doc

docs/explanation/features/repository-mirroring-2604.md
YanisaHS, line 1: Also - I think the sidebar should be 

- "Repository mirroring (pre-26.04)" and 
- "Repository mirroring (26.04+)

and their H1s can be

- "Repository mirroring (25.10 and earlier)"
- "Repository mirroring (26.04 LTS and later)"

still cross-linking to each other as a note

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+## Core concepts
```
YanisaHS:
```suggestion
## Repository mirroring concepts
```

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Repository mirroring in Landscape is built around four core entities: **mirrors**, **local repositories**, **publications**, and **publication targets**. Understanding these concepts and how they relate to each other is crucial to using this feature correctly.
```
YanisaHS, line 26:
```suggestion
Repository mirroring in Landscape is based on these main concepts: 

- **Mirrors**: Local copies of upstream Debian repositories.
- **Local repositories**: Repositories that contain `.deb` packages you provide
- **Publications**: Published versions of mirrors or local repositories.
- **Publication targets**: Storage locations where published repositories are written.

[continue more - how do these concepts relate?]
```
(reword as needed to match the content)

And what about repository profiles? Should there be another section on that? (not sure) - it could also be here, and then in the section below as "(Optional) Repository profiles"

--

But in the [continue] part I left - I think this section could use more connective pieces, so it's not like we're just providing definitions for the user to figure out how they relate. The previous explanation uses diagrams, which would be useful here to show the flow of where packages are

Can you connect these concepts via a (ideally mermaid) diagram?

And, in addition (or instead of), is there anything else you can say about how these concepts relate? (e.g., some of those are for the actual repository contents, and some is for how that content is made available to clients, right?)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Mirrors
+
+A mirror is a local copy of an upstream Debian repository (for example, `archive.ubuntu.com`). When you create a mirror, you specify:
+
+- **Archive root:** The upstream URL to mirror from (e.g. `http://archive.ubuntu.com/ubuntu/`)
```
YanisaHS:
```suggestion
- **Archive root:** The upstream repository URL to mirror from (e.g. `http://archive.ubuntu.com/ubuntu/`)
```
(if that's correct)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+- **Distribution:** The repository suite to mirror. This corresponds to what the Ubuntu archive calls a series and pocket combination. For example, `noble` (the release pocket of Ubuntu 24.04 LTS), `noble-updates` (the updates pocket), or `noble-security` (the security pocket)
+- **Components:** The categories of packages to include. Upstream Ubuntu repositories typically use `main`, `restricted`, `universe`, and `multiverse`
```
YanisaHS: "typically use" ? Are these not the ones that are always used?

YanisaHS: Is there a reason you chose not to copy the longer definitions from https://documentation.ubuntu.com/landscape/explanation/features/repository-mirroring/?

Not that it needs to be word-for-word, but remember that new users to 26.04 might not know what any of this stuff means (distributions and components). I think the relevant parts of those definitions should be carried over

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+- **Architectures:** The CPU architectures to mirror (e.g. `amd64`, `arm64`)
+- **Filter (optional):** A package query expression to select a subset of packages from the upstream repository, optionally including their dependencies
+
+After creating a mirror, you **sync** it to download packages from the upstream repository. Syncing is a long-running operation that runs in the background. You can sync a mirror repeatedly to pull in the latest packages from upstream.
+
+### Local repositories
+
+Local repositories let you host your own `.deb` packages that aren't sourced from an upstream mirror. You can import packages into a local repository by providing a URL to a `.deb` file or archive of `.deb` files. This is useful for distributing internally-built software or third-party packages that aren't available in any upstream repository.
```
YanisaHS:
```suggestion
Local repositories let you host your own `.deb` packages that aren't sourced from an upstream mirror. You can use a local repository for distributing internally-built software or third-party packages that aren't available from an upstream repository you mirror. You add packages to the local repository, then publish the repository so that client machines can use it as an APT source.
```
re-word as you want - suggestion is to organize it cleaner to how someone would think about it, so now it's (1) what it is (2) why use it (3) how to use it

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Each local repository has a default distribution and component, which are used when packages are published.
+
+### Publications
+
+A publication connects a **source** (a mirror or local repository) to a **publication target** (a storage backend). It defines *how* your repository is made available to clients by configuring:
```
YanisaHS:
```suggestion
Publications make the mirror or local repository available to your client machines. A publication connects a **source** (a mirror or local repository) to a **publication target** (a storage backend). It defines *how* your repository is made available to clients by configuring:
```
Or something like that? If that's correct?

I'm trying to include more connective pieces in so we're not just defining the topics - i.e., how does this concept relate to the packages and getting them on the clients?

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+- **Source:** Which mirror or local repository to publish
```
YanisaHS:
```suggestion
- **Source:** The mirror or local repository to publish
```

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+- **Publication target:** Where to publish (see below)
```
YanisaHS: No "see belows" - link to the section

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+- **Distribution:** The suite name clients will use in their APT configuration
+- **Signing key:** A private GPG key used to sign the published repository metadata
+- **Metadata options:** Labels, origins, architectures, and compression settings
```
YanisaHS: Anything else to add here? It's not clear to me what "metadata options" is

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Publishing a publication is a long-running operation. When you publish, the service takes a point-in-time snapshot of the source mirror or local repository and writes the resulting APT repository structure to the publication target.
```
YanisaHS:
```suggestion
When you publish, Landscape creates a point-in-time snapshot of the source mirror or local repository and writes the resulting APT repository structure to the publication target. Client machines can then be configured to use that published repository.
```
(again, if that's correct)

Trying to connect the concepts to what's actually happening to the packages / clients

Also - the first part about it being a long-running task doesn't fit here, I'm not sure why it's mentioned in general? I'll move it down though

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+You can create multiple publications from the same source, each going to a different target or using different settings. This lets you, for example, publish the same mirror to both a local filesystem for internal use and an S3 bucket for remote clients.
+
```
YanisaHS, line 64:
```suggestion
Publishing a publication is a long-running operation that [why is this mentioned? because it takes a long time so users should be prepared?]
```
I don't get why this is mentioned? If it is because it takes forever, then it can be omitted actually and just included in the how-to guide

But if it's something else - moving down here would likely be better anyway. (not sure why it being a "long-running operation" is here though)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+- **Source:** Which mirror or local repository to publish
+- **Publication target:** Where to publish (see below)
+- **Distribution:** The suite name clients will use in their APT configuration
+- **Signing key:** A private GPG key used to sign the published repository metadata
+- **Metadata options:** Labels, origins, architectures, and compression settings
+
+Publishing a publication is a long-running operation. When you publish, the service takes a point-in-time snapshot of the source mirror or local repository and writes the resulting APT repository structure to the publication target.
+
+You can create multiple publications from the same source, each going to a different target or using different settings. This lets you, for example, publish the same mirror to both a local filesystem for internal use and an S3 bucket for remote clients.
```
YanisaHS:
```suggestion
You can create multiple publications from the same source, each going to a different target or using different settings. This lets you, for example, publish the same mirror to both a local filesystem for internal use and another target for remote clients.
```
Or "another target (such as an S3 bucket) for..." ?

I wanted to add the term "target" back in to keep things building in knowledge for the user

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Publication targets
+
+A publication target is a named storage destination where published repositories are written. Landscape supports three types of publication target:
```
YanisaHS:
```suggestion
A publication target is the storage location where Landscape writes a published repository. Landscape supports the following types of publication targets:
```

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+- **Filesystem:** A directory on the local filesystem. You can configure whether files are hardlinked, symlinked, or copied.
+- **S3:** An Amazon S3 bucket or S3-compatible object store (such as MinIO). Configuration includes access credentials, region, bucket name, prefix, and storage class.
+- **Swift:** An OpenStack Swift container. Configuration includes authentication URL, credentials, container name, and prefix.
```
YanisaHS: Are the configuration parts necessary? I'm not sure, they just feel like reference material to me and introduce a lot of concepts the reader needs to sift through that might not be necessary to understand how repository mirroring actually works

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Publication targets are independent of the repositories themselves. You define them once and can reuse them across multiple publications.
```
YanisaHS:
```suggestion
Publication targets are separate from mirrors and local repositories. You can define a target once and reuse it for multiple publications.
```
If that's still in-line with what you had? This reads clearer to me, more similar terms you already used in the guide

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
```
YanisaHS, line 64: Should there be a section for repository profiles? (maybe not, idk)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+## Mapping to Ubuntu archive terminology
```
YanisaHS: I think this whole section should be cut out of here actually, and the components and distribution portions should be in that above relevant section instead (I commented it above already)

Because how it reads now - essentially a new 26.04 Landscape user needs to learn/be exposed to old terminology they won't need again to go through this explanation, when this explanation should be a "fresh" take on how repository mirroring works in Landscape, same as the other guide that's for 25.10 and earlier

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+**Components** categorise packages within a distribution:
+
+- **main:** Packages directly maintained by the repository owner (Canonical, for Ubuntu)
+- **restricted:** Proprietary packages and drivers
+- **universe:** Community-maintained packages
+- **multiverse:** Community-maintained packages with additional restrictions
+
+When you create a mirror, you select which components to include. A single mirror can include any combination of components.
+
+## Filters
```
YanisaHS: These sections on Filters and Signature-preserving mirrors feel like they come out of nowhere?

I'm trying to figure out where to put them though - maybe they should be as sub-sections under "Mirrors"? Instead of "Filters", this could be re-frames as "Filtered mirrors"?

What are your thoughts?

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Mirrors support package-level filtering using a query expression language. Filters let you select a specific subset of packages from the upstream repository, replacing the role that pull pockets played in earlier versions of Landscape.
```
YanisaHS: In general, it should be clearer what it actually is in the first sentence, e.g., "A filtered mirror includes only part of an upstream repository." or (if you're not switching it to "filtered mirrors") "You can use filters to include only part of an upstream repository in your mirror.", etc ..

Also - don't mention pull pockets here - this one should be about how repository mirroring works _now_

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+Filters are applied at sync time. If you need to distribute different subsets of packages to different groups of machines, you can create multiple filtered mirrors from the same upstream repository and publish each one separately.
+
+```{note}
+Filters cannot be used on signature-preserving mirrors, since filtering could invalidate the upstream repository's original GPG signatures.
+```
+
+## Signature-preserving mirrors
+
+When you create a mirror, you can enable **signature preservation**. A signature-preserving mirror passes through the upstream repository's original GPG signatures without re-signing, which is useful when you want clients to verify packages directly against the upstream key.
```
YanisaHS: Same here - the "what is this?" should be super clear upfront in the first sentence, e.g., "A signature-preserving mirror keeps the upstream repository signatures instead of creating new signed repository metadata." (or whatever)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+This mode has restrictions: you cannot apply filters to a signature-preserving mirror, and syncing does not occur until publication time. The mirror is treated as a direct pass-through of the upstream repository.
+
+## An example mirroring workflow
+
+The following example illustrates how a Landscape administrator could use repository mirroring to manage package distribution:
+
```
YanisaHS, line 136: of course I'm extremely biased because I wrote it 😇 BUT personally I prefer the format of [the previous repo mirroring example](https://documentation.ubuntu.com/landscape/explanation/features/repository-mirroring/#an-example-repository-mirroring-process). This is because this section isn't actually a "how-to" guide (with named steps, etc.), but just walking through an example scenario. So personally I'd suggest removing the step names and closer aligning to the previous one

Also, yours doesn't describe a "real" scenario - e.g., why is this hypothetical admin making two mirrors? It doesn't need to be the test/prod example - I just thought that one is easiest to think about - but there should still be a real scenario in mind here

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+   - One for `noble` (the release pocket) with components `main` and `universe`
+   - One for `noble-security` (the security pocket) with the same components
+1. **Sync the mirrors.** Both mirrors are synced to download all matching packages from the upstream repository.
+1. **Create a filtered mirror for testing.** They create a third mirror of `noble-updates` with a filter that selects only packages relevant to their application stack.
+1. **Publish.** They create publications for each mirror, all targeting the filesystem publication target, and publish them. Clients can now be configured to use a local server (e.g., Nginx or Apache Server) as their APT source.
+1. **Iterate.** When new upstream updates are available, the administrator syncs the mirrors again and re-publishes to make the updated packages available.
+
+## GPG keys
+
+GPG keys serve two distinct purposes in Landscape repository mirroring: **verifying** upstream repositories and **signing** published repositories.
```
YanisaHS:
```suggestion
GPG keys are used with repository mirroring in Landscape to establish trust in the mirror and verify the packages originated from a trusted source. The GPG keys are used for two purposes:

- Verifying packages from upstream repositories
- Signing published repositories for client machines
```
re-word as necessary - trying to add more general information upfront

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Verification keys (mirrors)
+
+When Landscape downloads packages from an upstream repository, it verifies the repository's GPG signatures to ensure the packages are authentic and unmodified. This requires the **public GPG key** of the upstream repository.
```
YanisaHS:
```suggestion
When Landscape downloads packages from an upstream repository, it verifies that the repository's metadata is signed by a trusted key to ensure the packages are authentic and unmodified. This requires the **public GPG key** of the upstream repository.
```
(assuming that's correct)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+For Ubuntu repositories, the public GPG keys (Ubuntu Archive 2012 and 2018 signing keys) are built into the debarchive service and configured automatically. No manual setup is required.
```
YanisaHS:
```suggestion
For Ubuntu repositories, the public GPG keys are built into Landscape and configured automatically. No manual setup is required.
```
The user doesn't need to know about debarchive here - they're just experiencing Landscape as the whole product, not it's individual components

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+1. **Sync the mirrors.** Both mirrors are synced to download all matching packages from the upstream repository.
+1. **Create a filtered mirror for testing.** They create a third mirror of `noble-updates` with a filter that selects only packages relevant to their application stack.
+1. **Publish.** They create publications for each mirror, all targeting the filesystem publication target, and publish them. Clients can now be configured to use a local server (e.g., Nginx or Apache Server) as their APT source.
+1. **Iterate.** When new upstream updates are available, the administrator syncs the mirrors again and re-publishes to make the updated packages available.
+
+## GPG keys
+
+GPG keys serve two distinct purposes in Landscape repository mirroring: **verifying** upstream repositories and **signing** published repositories.
+
+### Verification keys (mirrors)
```
YanisaHS: I don't know why this says "(mirrors)" - can it change to "(for upstream repositories)" ? Or just remove it

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+For Ubuntu repositories, the public GPG keys (Ubuntu Archive 2012 and 2018 signing keys) are built into the debarchive service and configured automatically. No manual setup is required.
+
+If you're mirroring a third-party repository, you need to provide its public GPG key when creating the mirror:
+
+1. Obtain the third party's public GPG key
+1. Ensure it's in ASCII-armored format
+1. Provide it when creating the mirror in Landscape
+
+### Signing keys (publications)
```
YanisaHS: also not sure why this says "publications" - can you remove it or change to "for client machines" (or whatever)?

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,153 @@
...
+
+When you publish a mirror or local repository, the debarchive service generates its own repository metadata (package indices, release files) for the published repository. A **private GPG key** is used to sign this metadata, so that client machines can verify the published repository is trustworthy.
```
YanisaHS:
```suggestion
When Landscape publishes a mirror or local repository, the published repository metadata must be signed. Client machines use the corresponding public key to verify that the published repository is trusted.

The signing key has two parts:

- Private key: Used by Landscape to sign the published repository metadata.
- Public key: Distributed to client machines so they can verify the published repository.
```
(reword as necessary, not sure if that's all correct) but can you build off of that framing? I found this section confusing, not presented in a scaffolded way (or at least not one that I understood :D)

YanisaHS (changes requested):
Ok I'm done now, have fun 🎊

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yes, I've already removed it.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Nice. I'll work it in there.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Doing it.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Adding it.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: I decided to add a bullet point regarding repository profiles, plus the mermaid.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Some non-ubuntu ones have different ones. For example, debian has `main`, `contrib`, and `non-free`.

Perfect5th: But I guess since I'm referring specifically to Ubuntu here, I can remove the work "typically".

Perfect5th: They seemed too wordy to me. Would it be okay to just link https://documentation.ubuntu.com/project/how-ubuntu-is-made/concepts/package-archive/ ?

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yes, pretty correct.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: I can expand all the currently existing options, sure.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yes, that's correct. Definitely okay to remove references to LRO.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: I've removed references to LROs, as that's a implementation detail and not really relevant. I think user's will understand that some things take awhile.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Fair enough. Plus, the user interface makes it at least somewhat clear. I'll remove some detail.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yes, lgtm!

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: I've mentioned them earlier and added them to the new diagram. I think that should be enough.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yup, I've removed it.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: I like that idea. They are "sub-types" of mirrors, in a way, so that fits. I'll move the sections and rename them.

Perfect5th: I've also swapped the order so that the note in filtered mirrors about signature-preserving mirrors comes after the reader reads about the existence of signature-preserving mirrors.

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: It was to differentiate it from "Signing keys (publications)", but the body of the sections also does a good job of that, so I'll just delete the parentheticals.

Perfect5th:
Ok, I think I've addressed all of the feedback.

Biggest change is the example mirroring workflow section. I tried to ~~rip-off~~ take inspiration from prior art, though I may have overcomplicated the diagram.

Essentially the flow I envision is using syncs and separate publication targets to "release" packages to clients with specific repository profiles.

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,164 @@
...
+- **Repository profiles:** Configurations that can apply published repositories to client machines.
+
+```{mermaid}
+graph TD
+    A[Mirror] -->|Source for| C[Publication]
+    B[Local Repository] -->|Source for| C[Publication]
+    C[Publication] -->|Publishes to| D[Publication Target]
+    D[Publication Target] -->|Serves repository to| E[Client Machines]
+    F[Repository Profile] -->|Applies APT sources to| E[Client Machines]
+```
```
YanisaHS, line 44:
```suggestion
```{mermaid}
graph TD
    A[Mirror] -->|Source for| C[Publication]
    B[Local repository] -->|Source for| C
    C -->|Publishes to| D[Publication target]

    D -->|Serves repository to| E[Client machines]

    F[Repository profile] -.->|Configures managed clients to use the APT source| E
    D -.->|Repository URL used in profile| F
```

Image


What are your thoughts on something like this? If it's correct. I thought the publication target is still needed for the repository profile, which is what I'm trying to solve in this suggestion.

I'm fine with yours otherwise though if it's correct!

docs/explanation/features/repository-mirroring-2604.md
YanisaHS: Ok! Yea - can you add something in addition to the shorter definitions you already have that links here? Like "For more details see ...."

docs/explanation/features/repository-mirroring-2604.md
YanisaHS: Is this hierarchy as intended? That filtered mirrors is a sub-set of signature-preserving mirrors? (I was expecting signature-preserving and filtered mirrors to be at the same level, both as a subset of Mirrors)
image

YanisaHS (changes requested):

docs/explanation/features/repository-mirroring-2604.md
Perfect5th: Yes, that works for me, thanks!

Perfect5th:
>Ok! Yea - can you add something in addition to the shorter definitions you already have that links here? Like "For more details see ...."

I made the names of each concept links, rather than adding additional text. Let me know if that seems ok.

docs/explanation/features/repository-mirroring-2604.md
YanisaHS: @Perfect5th (from your comment below) I like what you did with that part - but you missed this change too (in the **Mirrors** section - Distribution/components/ ....)

docs/explanation/features/repository-mirroring-2604.md
```diff
@@ -0,0 +1,177 @@
...
+- **Source:** The mirror or local repository to publish
+- **{ref}`Publication target<explanation-repo-mirroring-2604-publication-targets>`:** Where to publish
+- **Distribution:** The suite name clients will use in their APT configuration
+- **Signing key:** A private GPG key used to sign the published repository metadata
+- **Metadata options:**
+  - The values of the Label and Origin fields in the published repository's Release file
+  - Which architectures to include in the published repository
+  - Whether to provide hash index files
+  - Settings for the ButAutomaticUpgrades and NotAutomatic fields in the Release file
+  - Settings for using compression and generating content index files
```
YanisaHS, line 98:
```suggestion
- **Metadata options:**
  - The values of the `Label` and `Origin` fields in the published repository's `Release` file
  - Which architectures to include in the published repository
  - Whether to provide hash index files
  - Settings for the `ButAutomaticUpgrades` and `NotAutomatic` fields in the `Release` file
  - Settings for using compression and generating content index files
```

YanisaHS (commented):
There's another comment I made above too but otherwise LGTM

Perfect5th:
> There's another comment I made above too but otherwise LGTM

Not sure which other comment you're referring to, sorry.

YanisaHS:
@Perfect5th This one: https://github.com/canonical/landscape-documentation/pull/225#discussion_r3311737184, linking the Ubuntu package archive docs that describe each of the pieces in a mirror

Perfect5th:
Okay I think I've addressed it now.

YanisaHS (approved):
✨

PR #226, opened 2026-05-21
fix: database name should be set manually for debarchive

Perfect5th:
Adding the dbname for debarchive to the service.conf causes issues with landscape-server, as it doesn't recognize the key and does not tolerate extra keys in its config.

Instead this value should be set manually using `snap set`.


jmucc (approved):
LGTM

PR #227, opened 2026-05-25
Add reprepro to debarchive migration instructions

Perfect5th:
Testing the instructions is a little involved:

1. install landscape-server-quickstart from the 25.10 PPA
2. create at least one mirror sync pocket (I used jammy-backports universe for an around ~120 package repo), at least one pull pocket, and at least one local repo
3. upload at least one package to the local repo… this can be annoying to do. I built landscape-client with 
```Makefile
override_dh_strip:
       dh_strip --no-automatic-dbgsym
```
added to the `debian/rules` file and `dpkg-buildpackage -b -us -uc`
You can follow the [upload instructions in the docs](https://documentation.ubuntu.com/landscape/how-to-guides/repository-mirrors/manage-repositories-with-the-api/#upload-pockets).
4. `snap install landscape-debarchive` (see [the setup instructions](https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/debarchive-repository-management/))
5. follow the instructions, get a JWT from the brower's localStorage or log in via curl, use `localhost:8100` on the machine as `API_BASE`.
6. upgrade to 26.04 - you may need to delete your repository distribution and pockets first.
7. validate that your mirrors and local repos are there and they work.


docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,453 @@
...
+curl -X POST "$API_BASE/locals/<LOCAL_ID>:importPackages" \
+  -H "Authorization: Bearer $JWT" \
+  -H "Content-Type: application/json" \
+  -d '{"url": "file:///path/to/my-package_1.0-1_amd64.deb"}'
+```
+
+This is a long-running operation. Poll the returned operation for completion before importing the next package.
+
+```{note}
+When using `file://` URLs, the path must be accessible to the `landscape-debarchive` snap. Because snap confinement restricts filesystem access, `file://` URLs pointing to arbitrary paths outside the snap's data directory will not work. Packages must instead be served over HTTP — either by the Landscape Server itself, or by a temporary HTTP server.
```
jmucc, line 334: You're missing the closing ``` for the note

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,453 @@
...
+```
+
+Alternatively, start a temporary HTTP server:
+
+```bash
+python3 -m http.server 8080 --directory /path/to/packages
+```
+
+Then reference packages as `http://localhost:8080/my-package_1.0-1_amd64.deb`.
+```
```
jmucc:
```suggestion
```

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,453 @@
...
+---
+
+(how-to-migrate-repository-mirrors-to-debarchive)=
+# How to migrate repository mirrors to Deb Archive
+
+This guide describes how to migrate your existing reprepro-managed repository mirrors from Landscape 25.10 (and earlier) to the new Deb Archive service introduced in Landscape 26.04 LTS.
+
+```{note}
+This guide assumes you have already:
+- Upgraded to Landscape 26.04 LTS ({ref}`how-to-upgrade-to-26-04-lts`)
```
jmucc: Do we upgrade before starting this guide? It seems like we need to refer back to the reprepro backed deployment to configure Deb Archive, and wouldn't that information be lost if we upgraded? But we need to upgrade to set up  Deb Archive, so how did you intend for the user to have reprepro config available while configuring Deb Archive?

jmucc (commented):
I have a couple comments and wondering at what point to actually upgrade but I'll start setting up the 25.10 deployment in the meantime

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Ah right this is a good point.

The upgrade is blocked anyways if the old distribution records still exist in the database. I think a better workflow would be to:

1. install and configure `landscape-debarchive`
2. perform the migration steps
3. upgrade landscape-server

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: So that's where those got to…

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
jmucc: Ah but we can't configure `landscape-debarchive` before upgrading. All services crash if we add debarchive stuff to the `service.conf`. 

```
Extra inputs are not permitted [type=extra_forbidden, input_value='landscape-standalone-debarchive', input_type=str]
May 27 20:31:07 landscape-test log-system-event[76621]:     For further information visit https://errors.pydantic.dev/2.4/v/extra_forbidden
```

I am going to snapshot my 25.10 installation with all the reprepro stuff and take notes of everything needed in the guide and see if that works. Let me know if I'm missing something though.

Perfect5th: #226 - uhuh don't set it in the service.conf.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+Architectures: amd64
+SignWith: <KEY_FINGERPRINT>
+```
+
+The `Update:` field indicates a sync (mirror) pocket. Stanzas with a `Pull:` field are pull pockets. Stanzas without either are upload pockets.
+
+To see what packages are currently in a pocket:
+
+```bash
+reprepro -b /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION> list noble-release
```
jmucc: nit: Use `<Codename>` instead of `noble-release`?

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+```bash
+# Example: find all .deb files for the staging pocket's component
+find /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION>/pool/main/ -name "*.deb"
+```
+
+Alternatively, if the packages are served by the existing Landscape repository web server, they may already be accessible at:
+
+```
+https://$LANDSCAPE_FQDN/repository/standalone/<DISTRIBUTION>/pool/
+```
```
jmucc, line 301: This didn't work for me though it only says `"may"`. I just got a Landscape 404 page.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+```bash
+reprepro -b /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION> list noble-staging
+```
+
+The pool directory containing the actual `.deb` files is at:
+
+```
+/var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION>/pool/
+```
+
```
jmucc: This is true, but for me it is also listing all the packages from my other pockets. It's probably because I did `noble-backports universe` for all my pockets but raising in case you think there should be a change here.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+```bash
+python3 -m http.server 8080 --directory /path/to/packages
+```
+
+Then reference packages as `http://localhost:8080/my-package_1.0-1_amd64.deb`.
+
+```{important}
+The reprepro pool directory is shared across all distributions in the same reprepro tree. Scanning or archiving the entire pool will include packages from sync mirror and pull pockets, not just your upload pocket. Use one of the approaches below to import only the packages that belong to this specific pocket.
+```
```
jmucc, line 353: I think you can move this note up and link to this section or something similar. Personally I had this exact same question but I only came across this note as I was about the copy over only the packages I cared about.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
jmucc: Okay there's a note for this later, but I think the info could be useful here.

jmucc (commented):
Almost done validating. Feel free to wait until I finish the guide to address the comments but I'm not sure when I'll get back to this.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+
+#### Script imports from reprepro list
+
+Use `reprepro list` to enumerate only the packages belonging to this pocket, then locate each `.deb` file and import it via HTTP. Set `REPO_HTTP` to the base URL from which the repository is served:
+
+```bash
+#!/bin/bash
```
jmucc, line 360: Script seems to work. We discussed live and some basic error handling would be nice to have.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+### 5. Verify imported packages
+
+Once all imports have completed, confirm the packages are present in the local repository:
+
+```bash
+curl -X GET "$API_BASE/locals/<LOCAL_ID>/packages" \
+  -H "Authorization: Bearer $JWT"
+```
```
jmucc, line 413: This is really nice, I think it would be useful to have for the other sections.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
jmucc: Approved https://github.com/canonical/landscape-documentation/pull/226, as we discussed it might make the most sense to move the step about upgrading to  after `Migrate Upload Pockets`

jmucc (changes requested):
Almost there, I'll give one more pass over after the improvements we discussed.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
jmucc: I think this worked when I swapped to `http` instead of `https`. I think because my browser doesn't trust the cert so it may not be an issue for deployments with actual domains.

Perfect5th: I think the default apache setup only serves this via HTTP. I will change it.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,453 @@
...
+This is a long-running operation. Poll the returned operation for completion before importing the next package.
+
+```{note}
+When using `file://` URLs, the path must be accessible to the `landscape-debarchive` snap. Because snap confinement restricts filesystem access, `file://` URLs pointing to arbitrary paths outside the snap's data directory will not work. Packages must instead be served over HTTP — either by the Landscape Server itself, or by a temporary HTTP server.
+
+If the Landscape Server is already serving the repository, packages are accessible at:
+
+```
+http://$LANDSCAPE_FQDN/repository/standalone/<DISTRIBUTION>/pool/m/my-package/my-package_1.0-1_amd64.deb
+```
```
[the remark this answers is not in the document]
Perfect5th: This has been fixed. It should only be HTTP.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I'm replicating the note further up and slightly adjusted the wording.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I'm adding `set -euo pipefail` to the script, which will surface errors better.

Would be nice to have an independent test of it 😉 .

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,452 @@
...
+```bash
+curl -X GET "$API_BASE/locals/<LOCAL_ID>/packages" \
+  -H "Authorization: Bearer $JWT"
+```
+
+## Preserve the exact state of a sync mirror
+
+If you need to preserve the precise package versions currently in a sync mirror, rather than re-syncing from upstream (which may have newer versions), treat it as an upload pocket and import the packages into a local repository. Use the steps above for migrating upload pockets, but use the sync mirror pocket instead.
+
+## Publish the migrated repositories
```
Perfect5th: We discussed doing these steps in the Landscape UI, after upgrading to 26.04. I think I'll add the upgrade instructions here and have these reference using the UI.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
jmucc: Yup the script still works.

jmucc (commented):
I finished going through the instructions. Couple things.

- I think the note about upgrading to 26.04 at the start should be removed. Also, when linking to the upgrade guide later it might be worth noting that the debarchive steps should already completed. However, for me on quickstart I had to troubleshoot to get the debarchive service healthy again after upgrading. My final solution was to make the  landscape user the owner of all tables in the debarchive database and then do the same for landscape-maintenance, but I would think they just needed ownership rights or something. Did you test on quickstart and if so did you run into this?
- I tested publication targets in the UI.  I only tested filesystem publications but I was able to find my packages in `/var/snap/landscape-debarchive/common/publications/<HARDLINK_PATH>/pool/universe`. It wasn't explictly stated they would be there but I was able to figure that out. I know S3 and Swift are preferred for users to use so it might not be worth the trouble to clarify that, your call.

Perfect5th:
> I think the note about upgrading to 26.04 at the start should be removed.

Can do.

> Also, when linking to the upgrade guide later it might be worth noting that the debarchive steps should already completed. However, for me on quickstart I had to troubleshoot to get the debarchive service healthy again after upgrading. My final solution was to make the landscape user the owner of all tables in the debarchive database and then do the same for landscape-maintenance, but I would think they just needed ownership rights or something. Did you test on quickstart and if so did you run into this?

I also tested on quickstart and I don't recall running into this problem. Can you send me the logs, if you still have them? It sounds easy enough to replicate, as it probably doesn't require any of the actual repo migration steps - just an upgrade _after_ setting up debarchive.

> I tested publication targets in the UI.  I only tested filesystem publications but I was able to find my packages in `/var/snap/landscape-debarchive/common/publications/<HARDLINK_PATH>/pool/universe`. It wasn't explictly stated they would be there but I was able to figure that out. I know S3 and Swift are preferred for users to use so it might not be worth the trouble to clarify that, your call.

I think this is made clear in other documentation regarding debarchive, so I don't think I'll cover it here.

jmucc:
Here's what I saw in postgres logs, also saw this for `landscape@landscape-standalone-debarchive`. Can't seem to dig up the snap logs but it wrapped this error.

```
2026-05-29 14:42:55.569 UTC [72249] landscape_maintenance@landscape-standalone-debarchive ERROR:  must be owner of table checksum_info_aptly
2026-05-29 14:42:55.569 UTC [72249] landscape_maintenance@landscape-standalone-debarchive STATEMENT:  -- Relational schema for debarchive database
        -- This includes tables required for Aptly data,
        -- as well as any additional tables required for debarchive-specific data.
        -- Comments indicate which tables are for Aptly and which are for debarchive.
        
        -- Relational schema for storing mirrors for Aptly.
        CREATE TABLE IF NOT EXISTS mirror_aptly (
            uuid UUID PRIMARY KEY,
            name TEXT NOT NULL,
            archive_root TEXT NOT NULL,
            distribution TEXT NOT NULL,
            components TEXT[] DEFAULT '{}',
            architectures TEXT[] DEFAULT '{}',
            meta JSONB DEFAULT jsonb('{}'),
            last_download_date TIMESTAMP WITH TIME ZONE,
            release_files JSONB DEFAULT jsonb('{}'),
            filter TEXT,
            status INTEGER DEFAULT 0,
            worker_pid INTEGER DEFAULT 0,
            filter_with_deps BOOLEAN NOT NULL DEFAULT FALSE,
            skip_component_check BOOLEAN NOT NULL DEFAULT FALSE,
            skip_architecture_check BOOLEAN NOT NULL DEFAULT FALSE,
            download_sources BOOLEAN NOT NULL DEFAULT FALSE,
            download_udebs BOOLEAN NOT NULL DEFAULT FALSE,
            download_installer BOOLEAN NOT NULL DEFAULT FALSE
        );
        
...
```

Perfect5th:
When you upgraded landscape-server, how did you respond to the dpkg prompts asking whether you wanted to keep existing configuration files? This looks like an error related to re-generating or otherwise getting a new password in the service.conf.

I think it's a general upgrade rough-spot and is not specific to this guide.

jmucc:
I chose all the defaults (N), which should keep have kept the old configuration file. I agree it seems like something changed with the DB config before and after upgrade though. If you didn't run into this issue then it was probably an error on my part. Let me see if I can minimally repro it quick.

Perfect5th:
>Let me see if I can minimally repro it quick.

Thanks, that would be great.

jmucc:
Ok so I got it again.
1. Install 25.10 quickstart
2. Configure debarchive snap: https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/debarchive-repository-management/
3. See debarchive is active
4. Upgrade https://documentation.ubuntu.com/landscape/how-to-guides/upgrade/upgrade-landscape/
5. Debarchive is inactive: 2026-06-01T15:00:09Z landscape-debarchive.debarchive[23312]: {"level":"fatal","error":"ERROR: permission denied for schema public (SQLSTATE 42501)","time":"2026-06-01T15:00:09Z","message":"failed to create database tables"}
6. Rerun the db privileges command from the debarchive setup guide
7. Debarchive is still inactive: 2026-06-01T15:56:37Z landscape-debarchive.debarchive[140965]: {"level":"fatal","error":"ERROR: must be owner of table checksum_info_aptly (SQLSTATE 42501)","time":"2026-06-01T15:56:37Z","message":"failed to create database tables"}

jmucc:
> Ok so I got it again.
> 
> 1. Install 25.10 quickstart
> 2. Configure debarchive snap: https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/debarchive-repository-management/
> 3. See debarchive is active
> 4. Upgrade https://documentation.ubuntu.com/landscape/how-to-guides/upgrade/upgrade-landscape/
> 5. Debarchive is inactive: 2026-06-01T15:00:09Z landscape-debarchive.debarchive[23312]: {"level":"fatal","error":"ERROR: permission denied for schema public (SQLSTATE 42501)","time":"2026-06-01T15:00:09Z","message":"failed to create database tables"}
> 6. Rerun the db privileges command from the debarchive setup guide
> 7. Debarchive is still inactive: 2026-06-01T15:56:37Z landscape-debarchive.debarchive[140965]: {"level":"fatal","error":"ERROR: must be owner of table checksum_info_aptly (SQLSTATE 42501)","time":"2026-06-01T15:56:37Z","message":"failed to create database tables"}

I think I found the root cause, when upgrading from 25.10 to 26.04, the `set_user` extension becomes enabled. At this point `landscape-maintenance` gets created. I have this in my `service.conf` after upgrade

```
[schema]
store_password = ***********
store_user = landscape_maintenance
stores = main account-1 resource-1 package session
threads = 1
store_superuser = landscape_superuser
```

According to the [public docs](https://documentation.ubuntu.com/landscape/reference/config/service-conf/#the-schema-section), `landsape_maintenance` should escalate to `landscape_superuser`, but I am not sure if that's happening. Instead I get errors for `landscape_maintenance` having insufficient permissions.

I have found two workarounds

-  Replace `[schema]` section of the service.conf with the one from before the upgrade in `service.conf.old`.
- Running the below command in debarchive db (which makes landscape_maintenance the owner of all tables in landscape-standalone-debarchive). Also need the other `landscape_maintenance` command from the QS guide (that command is only relevant with the set_user extension enabled I think)
  - I'm not sure if this would break anything else in Landscape

```
landscape-standalone-debarchive=# DO $$ 
DECLARE 
    r RECORD;
BEGIN
    FOR r IN 
        SELECT schemaname, tablename 
        FROM pg_tables 
        WHERE schemaname NOT IN ('pg_catalog', 'information_schema')
    LOOP
        EXECUTE 'ALTER TABLE ' || quote_ident(r.schemaname) || '.' || quote_ident(r.tablename) || ' OWNER TO landscape_maintenance;';
    END LOOP;
END $$;
DO
```

TLDR: I'm getting database connection issues due to having the `set_user` extension turned on after the upgrade. I think the way forward may be to add some notes to the troubleshooting if you think the notes about db connectivity aren't enough. What's there already may be sufficient.

Perfect5th:
Good research, thanks.

I'm starting to think it might be best if debarchive has its own database user with its own set of permissions that will be unaffected by the upgrade.

This will require changes to https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/debarchive-repository-management/, not changes to this doc. Let's assume for now that's been done and this issue is "fixed". In that case, do we need any other changes to this doc? If not, I will ask for a review from @YanisaHS.

jmucc:
Yeah LGTM

Perfect5th:
>I'm starting to think it might be best if debarchive has its own database user with its own set of permissions that will be unaffected by the upgrade.

See #244

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+List all configured distributions and their pockets. Each distribution has its own configuration directory:
+
+```bash
+cat /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION>/conf/distributions
+```
+
+where `<DISTRIBUTION>` is the name of the distribution you created (e.g., `ubuntu`, `ubuntu-staging`).
+
+This file contains stanzas like:
```
YanisaHS:
```suggestion
This file contains sections like:
```
Or "entry", "block", or"configuration block"/"distribution block" ?

And can you change the other ones? "section" (or whichever term) is more plain English. Be consistent with the term you use (section/block/ .. )

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+|---|---|
+| **Sync (mirror) pocket** | Create a new mirror in Deb Archive pointing at the original upstream source and sync it |
+| **Pull pocket** | Create a new mirror with a `filter` matching your existing allowlist/blocklist |
+| **Upload pocket** | Create a local repository and import packages from the existing reprepro pool |
+
+```{important}
+If you need to preserve the **exact state** of a sync mirror (the precise set of package versions currently stored, rather than the latest upstream state), you should treat it as an upload pocket and import its packages into a new local repository instead.
+```
+
+## Prerequisites
```
YanisaHS, line 30: I'd consider "Prerequisites" as something you need to have in place and the instructions aren't within the guide. For example: https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/juju-ha-installation/#prerequisites

Can you change this heading? Maybe to "Before you begin" or just "Set environment variables"

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+    description: "Migrate repository mirrors from legacy reprepro-managed repositories (Landscape 25.10 and earlier) to the new debarchive service introduced in Landscape 26.04 LTS."
+---
+
+(how-to-migrate-repository-mirrors-to-debarchive)=
+# How to migrate repository mirrors to Deb Archive
+
+This guide describes how to migrate your existing reprepro-managed repository mirrors from Landscape 25.10 (and earlier) to the new Deb Archive service introduced in Landscape 26.04 LTS.
+
+```{note}
+This guide assumes you have already:
```
YanisaHS: You should take these out of a note and put them as a "Prerequisites" section, like https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/juju-ha-installation/#prerequisites

(after "Overview")

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+```bash
+curl -X GET "$API_BASE/locals/<LOCAL_ID>/packages" \
+  -H "Authorization: Bearer $JWT"
+```
+
+## Preserve the exact state of a sync mirror
+
+If you need to preserve the precise package versions currently in a sync mirror, rather than re-syncing from upstream (which may have newer versions), treat it as an upload pocket and import the packages into a local repository. Use the steps above for migrating upload pockets, but use the sync mirror pocket instead.
+
+## Upgrade to Landscape 26.04 LTS
```
YanisaHS, line 419: I'm not sure what the correct workflow should be, but this contradicts the prereqs listed at the top which says they should have already upgraded to 26.04?

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+Once you have confirmed that all packages are present in the new Deb Archive service, clean up the legacy reprepro Distribution records and then upgrade.
+
+### 1. Delete existing reprepro Distribution records
+
+In the Landscape web UI, navigate to **Repositories**. Each distribution (for example, `ubuntu` or `ubuntu-staging`) has a **Delete** button. Delete each distribution.
```
YanisaHS: all "web UI" should be "web portal"

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+- Installed and configured the `landscape-debarchive` snap ({ref}`how-to-debarchive-repository-management`)
+```
+
+## Overview
+
+In Landscape 25.10 and earlier, repository management was handled by an internal reprepro-based system that managed mirrors through distributions, series, and pockets. In Landscape 26.04 LTS, this system is replaced by the `debarchive` service, which provides a REST API for managing mirrors, local repositories, and publications.
+
+The migration strategy depends on the type of pocket you're migrating:
+
+| Legacy pocket type | Migration strategy |
```
YanisaHS:
```suggestion
| pre-26.04 pocket type | Migration strategy |
```
Thoughts? I wanted to be clearer about what we consider the "legacy" type

In general, I don't think we should use the term "legacy" throughout this guide since it's not a "legacy" system for the user (i.e., it's their current system until they migrate). Can you change those too to pre-26.04 or "existing system" ?

I'm open to other suggestions, but I want to move away from calling it "legacy", or even try to avoid "reprepro-based" in places where it doesn't make sense to explicitly call out reprepro

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+Once the sync is complete, confirm the packages are present in the mirror:
+
+```bash
+curl -X GET "$API_BASE/mirrors/<MIRROR_ID>/packages" \
+  -H "Authorization: Bearer $JWT"
+```
+
+### 5. Repeat for each sync pocket
+
+Repeat steps 2–4 for each sync pocket (`release`, `updates`, `security`, etc.).
```
YanisaHS: Is it correct that it's only steps 1-4 (i.e., they all have the same upstream source) .. ? Presumably no, but I don't actually know.

You could also more broadly say "Repeat the previous steps for each sync pocket"

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+```text
+Name: noble-release-staging
+From: noble-release
+Components: main
+Architectures: amd64
+FilterList: install noble-release-staging.list
+```
+
+The referenced filter list file (e.g., `noble-release-staging.list`) contains package name patterns, one per line. Examine the filter list:
```
YanisaHS: What are they examining it for? (I'd give a clearer suggestion, but I'm not sure what they're looking for)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+This guide describes how to migrate your existing reprepro-managed repository mirrors from Landscape 25.10 (and earlier) to the new Deb Archive service introduced in Landscape 26.04 LTS.
+
+```{note}
+This guide assumes you have already:
+- Upgraded to Landscape 26.04 LTS ({ref}`how-to-upgrade-to-26-04-lts`)
+- Installed and configured the `landscape-debarchive` snap ({ref}`how-to-debarchive-repository-management`)
+```
+
+## Overview
```
YanisaHS, line 12: It'd be helpful to link to https://documentation.ubuntu.com/landscape/explanation/features/repository-mirroring-2604/ in this section.

I'd suggest putting it at the end of the first paragraph or at the end of the full Overview, something like "For details on how repository mirroring works in 26.04, see XYZ"

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+```text
+nginx install
+curl install
+libssl3 install
+```
+
+### 2. Translate the filter to a Deb Archive filter expression
+
+Deb Archive mirrors support package filtering using an aptly-compatible filter syntax. Translate your allowlist into a filter expression:
```
YanisaHS: Is there another way you can describe this without saying "aptly-compatible"? I don't want to expose too much of the backend here if we can avoid it.

But if removing Aptly will just make it sound weird / ambiguous, then leave it worded as is and link to Aptly's docs or main product website

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+Example allowlist content:
+
+```text
+nginx install
+curl install
+libssl3 install
+```
+
+### 2. Translate the filter to a Deb Archive filter expression
```
YanisaHS:
```suggestion
### 2. Migrate allowlists and blocklists to filtered mirrors
```

In general - I found this section confusing. IMO it seems like they're still considered "allowlists" and "blocklists" in 26.04+. But you're trying to say that allow/block lists are now "filtered mirrors", right?

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+  -H "Authorization: Bearer $JWT"
+```
+
+### 5. Repeat for each sync pocket
+
+Repeat steps 2–4 for each sync pocket (`release`, `updates`, `security`, etc.).
+
+## Migrate pull pockets
+
+Pull pockets in the legacy system stage packages from another pocket using allowlist or blocklist filters. In Deb Archive, this is achieved by creating a mirror with a `filter` expression that replicates your allowlist or blocklist.
```
YanisaHS:
```suggestion
Pull pockets in the legacy system stage packages from another pocket using allowlist or blocklist filters. Starting in Landscape 26.04 LTS, this is achieved by creating a mirror with a `filter` expression that replicates your allowlist or blocklist, also referred to as a "filtered mirror".
```
Or "Starting in Landscape 26.04 LTS with Deb Archive" if you think it helps to be more explicit.

(1) The user is still experiencing these through _Landscape_, not directly through deb archive
(2) Keeping the terminology consistent - "filtered mirrors"

I think it's okay to call out Deb Archive sometimes in the guide when you are specifically referring to the actual service (e.g., if you're saying where the Deb Archive service runs), but here you're describing a general Landscape behavior

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
YanisaHS, line 1: I'm noticing that the examples don't have all the placeholders called out - only some are marked placeholders, and some use examples directly provided at the beginning of the step.

Can you go through the examples and make sure that's clear? They could just be resolved by clarifying it's an example call. Like instead of "create the mirror", change it to "create the mirror. Example call:" (or similar)

I like providing the examples for syntax rather than just placeholders everywhere, but how it is now, a user could copy it, change the placeholders, and not notice there's other "example" data inside it (components, filters, etc.)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+The `Update:` field indicates a sync (mirror) pocket. Stanzas with a `Pull:` field are pull pockets. Stanzas without either are upload pockets.
+
+To see what packages are currently in a pocket:
+
+```bash
+reprepro -b /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION> list <CODENAME>
+```
+
+Replace `<DISTRIBUTION>` with the appropriate distribution directory name and `<CODENAME>` with the codename of the pocket (e.g., `noble-updates`).
+
```
YanisaHS, line 93:
```suggestion

Use this information to decide which migration sections to follow. For example, if your deployment only has sync pockets, complete the sync-pocket migration steps and skip the pull-pocket and upload-pocket sections.
```

Right? Reword as needed, but something that ties this section together would be helpful, so it's clear what the user should do with this information.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+```bash
+curl -X POST "$API_BASE/mirrors/<MIRROR_ID>:sync" \
+  -H "Authorization: Bearer $JWT" \
+  -H "Content-Type: application/json"
+```
+
+### 5. Verify the filtered mirror
+
+Once the sync is complete, confirm the filtered packages are present in the mirror:
```
YanisaHS: What are they looking for to confirm? If it's obvious from the output, feel free to ignore this suggestion. That's just always a pet peeve of mine when a guide says to check that it worked with XYZ command, but it's not obvious from the output :)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+### 1. Identify the packages
+
+List the packages in the upload pocket:
+
+```bash
+reprepro -b /var/lib/landscape/landscape-repository/standalone/<DISTRIBUTION> list noble-staging
+```
+
+The pool directory containing the actual `.deb` files is at:
```
YanisaHS: Can you clarify what this step / the pool directory is for? (assuming the user won't know why they're looking at the pool directory)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,457 @@
...
+
+After publishing, update the repository profiles on your managed machines to point to the new Deb Archive publication URLs instead of the legacy repository paths.
+
+The legacy repository was served at:
+
+```
+deb https://$LANDSCAPE_FQDN/repository/standalone/ubuntu <codename>-<pocket> <components>
+```
+
+The new Deb Archive filesystem publications can be served from the configured published root path using the web server of your choice. Consult your publication target configuration for the exact URL.
```
YanisaHS: "the web server of your choice" should change, but I'm not sure what to. 

At this point, did they already choose the web server they're using when they did the upgrade? If yes:

"the web server you set up before" (or something like that - reference it's something they already did)

If they haven't, then

"a web server, such as nginx or Apache" (or something like that - give them a nudge)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I'll go with "Set environment variables"

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Right. We pivoted on this partway through.

It's much cleaner to upgrade at this point than before doing any of the steps. I'll remove the upgrade from the (new) prereqs section.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Seems alright to me. I'll replace mentions of "legacy" with "pre-26.04" throughout.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Yes, technically step 1 would be the same for each pocket from the same distribution, but it's also less confusing to just say "do all the steps" even if one is the same. I'll make that change too.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I think I can remove this "examine" directive. The example below, and the subsequent translation to the new format makes it pretty clear (imo) what they are looking for. Let me know if you disagree.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I think I can take it out.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Yes, that's what I'm trying to say. I like your suggestion here.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Makes sense. I'm okay with this change as-is.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Yes, I'll do this, but I'll leave it to last, as it's a bit more involving than the other changes you've commented on.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Yeah. I have re-worded slightly, but not by much.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: The call lists the packages in the mirror, and the user should read them to make sure they match what their filters were.

I'll see if I can make it a bit more obvious.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I'll flesh this out a little bit.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I like the "such as" part. We may end up adding a small web server to the deb archive snap, based on Field DC feedback, so I'll need to come back and change this a bit more, but for now I think it's fine.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: I've added the "example call"-style wording. I'm not sure it's completely clarifies it, but I think it helps a bit. FWIW, Joey didn't seem to have trouble identifying these as examples that should be edited.

Perfect5th: But maybe he's just really smart.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,467 @@
...
+|---|---|
+| **Sync (mirror) pocket** | Create a new mirror in Deb Archive pointing at the original upstream source and sync it |
+| **Pull pocket** | Create a new mirror with a `filter` matching your existing allowlist/blocklist |
+| **Upload pocket** | Create a local repository and import packages from the existing reprepro pool |
+
+```{important}
+If you need to preserve the **exact state** of a sync mirror (the precise set of package versions currently stored, rather than the latest upstream state), you should treat it as an upload pocket and import its packages into a new local repository instead.
+```
+
+For details on how repository mirroring works in 26.04, see {ref}` "Repository mirroring (26.04 LTS and later)"<explanation-repo-mirroring-2604>`.
```
YanisaHS: They syntax is off - can you remove the quotation marks? 

Image

docs/how-to-guides/upgrade/upgrade-to-26-04-lts.md
```diff
@@ -13,7 +13,7 @@ You must be running Ubuntu 26.04 LTS Resolute Raccoon, 24.04 Noble Numbat, or 22
 
 Note that Quickstart installations and upgrades to Landscape 26.04 LTS are not supported on Ubuntu 26.04. You must use Ubuntu 24.04 LTS or 22.04 LTS for Quickstart installations.
 
-If you use repository management in your Landscape deployment, we recommend waiting to upgrade to 26.04 LTS until the 26.04.1 point release (expected August 2026). A migration guide for bringing over repository mirrors from 24.04 LTS to 26.04 LTS will be published before the point release.
+If you use repository management in your Landscape deployment, we recommend waiting to upgrade to 26.04 LTS until the 26.04.1 point release (expected August 2026). See {ref}`how-to-migrate-repository-mirrors-to-debarchive` for instructions on bringing over repository mirrors from 25.10 and earlier to 26.04 LTS.
```
YanisaHS: If the strategy changed, then this guide should be more clear that it's for users without repository mirrors. (going to comment this in the other thread that's relevant, but pointing it out here as well)

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
YanisaHS: Ok, in that case, this should be mentioned in the full upgrade guide.

I think the easiest way to approach this (given what we currently have) is to add a new "If you use repository management" (or similar title) section below the intro / before [Begin your upgrade](https://documentation.ubuntu.com/landscape/how-to-guides/upgrade/upgrade-to-26-04-lts/#begin-your-upgrade) that clarifies if you're using Landscape to manage repositories, then you should start in the Migrate guide

The alternative would be to have two full upgrade guides - one for with repository management and one for without repository management that have duplicated content. I think the most practical approach given what we currently have is to do my first suggestion (add a section to the existing upgrade guide that links to this migrate guide), but still pointing this out if you disagree with the first option.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Prerequisites
+
+This guide assumes you have already:
+
+- Installed and configured the `landscape-debarchive` snap ({ref}`how-to-debarchive-repository-management`)
+
+## Set environment variables
+
+Set the following environment variables for use throughout this guide:
```
YanisaHS:
```suggestion
Set the following environment variables for use throughout this guide, using the FQDN of your Landscape deployment:
```
or something like that - just wanted to call it out so they don't accidentally take example.com

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
YanisaHS: Joey IS really smart :D but also I don't think most people would have issues, but some might and I want to reduce the friction if people are just copying codeblocks. It's not a big annoyance if you copy something and it fails, then you fix it, but still those annoyances add up ..... 😇

YanisaHS (changes requested):

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Oops, yeah.

docs/how-to-guides/upgrade/migrate-repository-mirrors-to-debarchive.md
Perfect5th: Sounds good. I'll also mention that it's a JWT for the REST API.

Perfect5th: Ugh also realized that "JWT token" is an "ATM Machine" situation, but also just "JWT" sounds awkward.

docs/how-to-guides/upgrade/upgrade-to-26-04-lts.md
Perfect5th: I can respond to this comment, but not the next one… thanks, Github.

Anyway I'll move this note into its own section and refer to the guide that. Hopefully that will make it clearer.

YanisaHS (approved):
LGTM as is, but reminder - does anything need to be connect to #244 ?

Also if that one's ready, can you find a reviewer (maybe me but you might want a technical review too?)

Perfect5th:
Only vaguely connected. I've asked for review (again) on #244 but I'd be comfortable merging this first.

jmucc (approved):
LGTM, taking 244 now

PR #233, opened 2026-05-28
feat: add tls and auth_mode to outbox conf reference

srunde3:


docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -1312,6 +1312,51 @@ Run the script `lsctl` to start the `landscape-server` daemons:
 sudo lsctl restart
 ```
 
+## Install and configure the Landscape Outbox
```
YanisaHS: Landscape 26.04 LTS+ only right? If yes, add that to this section

**Edit:** Actually, it might be better to change the heading, so it's "(Landscape 26.04+) Install and configure the Landscape Outbox"

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -1312,6 +1312,51 @@ Run the script `lsctl` to start the `landscape-server` daemons:
...
+
+The Landscape Outbox is a required component that interacts with the message broker and databases. Since the outbox runs as a snap under the `root` user, it requires its own copies of the client certificates for authentication.
```
YanisaHS: "Landscape Outbox" could link here: https://documentation.ubuntu.com/landscape/explanation/landscape/landscape-server-architecture/#outbox

Either with/without the crosslink is fine to me though

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -1312,6 +1312,51 @@ Run the script `lsctl` to start the `landscape-server` daemons:
...
+
+Copy the CA certificate and the RabbitMQ client certificates:
+
+```bash
+sudo cp /etc/ca-certificates.crt /root/snap/landscape-outbox/common/ca.crt
+sudo cp /etc/landscape/rabbitmq_client.pem /root/snap/landscape-outbox/common/rabbit.pem
+sudo cp /etc/landscape/rabbitmq_client.key /root/snap/landscape-outbox/common/rabbit.key
+```
+
+Ensure they are owned and readable by `root`:
```
YanisaHS:
```suggestion
Ensure the certificates are owned and readable by `root`:
```
"they" could refer to the certificates or the snap

docs/reference/config/outbox-configuration.md
YanisaHS, line 1: https://github.com/canonical/landscape-documentation/blob/main/docs/reference/config/outbox-configuration.md?plain=1#L21

So that part needs to change too, yes? The outbox doesn't read "all necessary database and broker configurations" from service.conf?

docs/reference/config/outbox-configuration.md
```diff
@@ -206,6 +206,22 @@ The outbox service connects to three PostgreSQL databases. Each database is conf
...
 
+#### `LANDSCAPE_BROKER_AUTH_MODE`
+
+- Purpose: Authentication mode to use with the AMQP broker. Must be one of `{plain,external}`.
+- Snap key: `landscape.broker.auth-mode`
+- Default: plain
+- Required: No
+- `service.conf`-supplied: No
+
+#### `LANDSCAPE_BROKER_TLS`
```
YanisaHS, line 217: Is there any connection to make here with the broker SSL one? (maybe not, just pointing it out) Doesn't `LANDSCAPE_BROKER_SSL_CA_CERT` already connect AMQP / TLS?

Here: https://github.com/canonical/landscape-documentation/blob/main/docs/reference/config/outbox-configuration.md?plain=1#L187

It says it uses AMQP, but in your doc now it's "whether to connect to AMQP"

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -1312,6 +1312,51 @@ Run the script `lsctl` to start the `landscape-server` daemons:
...
+
+The Landscape Outbox is a required component that interacts with the message broker and databases. Since the outbox runs as a snap under the `root` user, it requires its own copies of the client certificates for authentication.
+
+Install the `landscape-outbox` snap if not already installed:
+
+```bash
+sudo snap install landscape-outbox
+```
+
+Copy the CA certificate and the RabbitMQ client certificates:
```
YanisaHS: Copy these where? On the machine where `landscape-outbox` is installed? (the Server?)

YanisaHS (changes requested):

docs/reference/config/outbox-configuration.md
srunde3: My intent there was more to say "the outbox works out-of-the-box with quickstart". I think I'll revise it to say something more along the lines of "the outbox reads configuration from the Landscape service.conf and can populate many configuration marked **required** without them needing to be configured manually on the outbox".

docs/reference/config/outbox-configuration.md
srunde3: yes, good catch. This is now the only flag that controls TLS.

docs/reference/config/outbox-configuration.md
srunde3: I added a how-to for configuring the outbox; let me know if you think it's "overkill" and should be consolidated into existing docs. I wanted to get across why someone might want to override any configuration with `snap set ...`, and couldn't find a good spot for that in any existing docs.

YanisaHS: @srunde3 Regarding this outbox config - is this something a user might do outside of an installation process though? Or if they want to override settings, it would generally be during the install process?

Because if it's basically something someone just does during installation, I think it's better to find a way to integrate it in relevant docs, especially if it's not relevant for Quickstart. If it's the same steps throughout, note that you can use an include block like [this example](https://github.com/canonical/landscape-documentation/blob/main/docs/explanation/landscape/self-hosted-landscape.md?plain=1#L30) (the organization of those in our repo is a bit of a mess, but you get the idea). Presumably this information is relevant to other install guides aside from the DISA STIG one

You can also add a UI element [like the Customize COS Lite one in this section](https://documentation.ubuntu.com/observability/latest/tutorial/cos-lite-microk8s-sandbox/#deploy-cos-lite-using-terraform) ([source](https://github.com/canonical/observability-stack/blob/main/docs/tutorial/cos-lite-microk8s-sandbox.md?plain=1#L215))

LMK if you have questions or want to discuss this too!

srunde3: There are settings that they might want to reconfigure after installation, yeah. For example, they might add telemetry later, or they might want to change the maximum number of retries.

The point of the how-to was the separate the "here is how to configure the outbox in general" from "here is how you specifically need to configure the outbox for DISA STIG"

YanisaHS: @srunde3 ok! In that case, then yes I like including the guide, but it should be cross-linked to the DISA STIG and other relevant guides (like the upgrade to 26.04 one?) - "For more guidance on configuration , see ......", or whatever

And you should remove the installation parts from the config guide, make it more explicit that this is only relevant after the outbox has been installed, focus just on configuration stuff

YanisaHS (commented):
See: https://github.com/canonical/landscape-documentation/pull/233#discussion_r3325853681 - reviewing it too just to be explicit

The other changes LGTM!

YanisaHS (approved):
✨

PR #250, opened 2026-06-10
Apache samples serve debarchive repos on port 443

Perfect5th:
STIG requires HTTPS connections for serving packages.

This change adds the debarchive filesystem package locations to the sample apache conf under 443/https for disa-stig deployments.


Perfect5th:
Fixed some indentation, too.

HJK-X (approved):
LGTM

PR #257, opened 2026-06-16
fix: provide certs to manual TLS certs provider via action, not config

jansdhillon:
https://charmhub.io/manual-tls-certificates/docs/h-getting-started?channel=1/beta

1/beta is used because I tried using 1/stable and got a failure on install (for the manual TLS charm).

### DIFF
--- a/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
+++ b/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
@@ -17,16 +17,16 @@ This guide explains how to migrate from an older Landscape Server charm deployme
 
 The 26.04 beta version introduces significant architectural changes:
 
-| Aspect                   | Landscape 26.04 LTS beta+                                                                               | Pre-26.04                                                    |
-| ------------------------ | ------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------ |
-| **Load balancing**       | External HAProxy charm (`haproxy` at `2.8/edge`, `haproxy-route` interface)                             | External HAProxy charm (`reverseproxy` interface)            |
-| **PostgreSQL interface** | Modern `database` interface (PostgreSQL 14+)                                                            | Legacy `pgsql` interface (PostgreSQL 14)                     |
-| **PostgreSQL relation**  | `landscape-server:database` → `postgresql:database`                                                     | `landscape-server:db` → `postgresql:db-admin`                |
-| **RabbitMQ relation**    | `landscape-server:inbound-amqp` and `landscape-server:outbound-amqp` → `rabbitmq-server` (25.10+)       | `landscape-server:amqp` → `rabbitmq-server:amqp` (pre-25.10) |
-| **HAProxy relation**     | `landscape-server:*-haproxy-route` → `haproxy:haproxy-route` (8 route endpoints)                        | `landscape-server:website` → `haproxy:reverseproxy`          |
-| **TLS certificates**     | `haproxy:certificates` → TLS provider (e.g., `self-signed-certificates`, `lego`)                        | HAProxy self-signed or manual config                         |
-| **Access method**        | HAProxy unit IP or `root_url`                                                                           | HAProxy unit IP                                              |
-| **HA capabilities**      | HAProxy units for load balancing                                                                        | HAProxy units for load balancing                             |
+| Aspect                   | Landscape 26.04 LTS beta+                                                                         | Pre-26.04                                                    |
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -92,18 +92,71 @@ See the [lego charm documentation](https://charmhub.io/lego) for DNS-01 challeng
...
+juju run manual-tls-certificates/0 get-outstanding-certificate-requests --format=json | jq '.manual-tls-certificates/0.results.result' | jq '.[0].csr' > certs/client.csr
 ```
 
-See the [manual-tls-certificates charm documentation](https://charmhub.io/manual-tls-certificates) for details.
+```{tip}
+The outstanding certificate requests are grouped by the `relation-id`; if there are multiple requests (i.e., multiple consumers of the manaul TLS certificates), use `juju show-unit manual-tls-certificates/0` to identify correct the ID.
+```
+
+```{note}
+If the machine ID of the manual TLS certificates charm is not 0, adjust the above and following commands with the correct ID. Use `juju status` to identify the Juju machine ID of the manual TLS certificates charm.s
```
YanisaHS: Can you consolidate these into one note, just have a blank line in between? Just because I don't like how the stacked notes look :) Not sure if my suggestion block rendered correctly in the GH suggestions. (also you have some typos)

```suggestion
```{note}
The outstanding certificate requests are grouped by the `relation-id`; if there are multiple requests (i.e., multiple consumers of the manual TLS certificates), use `juju show-unit manual-tls-certificates/0` to identify the correct ID.

If the machine ID of the manual TLS certificates charm is not 0, adjust the above and following commands with the correct ID. Use `juju status` to identify the Juju machine ID of the manual TLS certificates charms.
```

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -92,18 +92,71 @@ See the [lego charm documentation](https://charmhub.io/lego) for DNS-01 challeng
...
+
+Now, generate a CA certificate and private key or ensure they are available locally. For example, generate a CA certificate and private key with OpenSSL:
+
+```sh
+mkdir certs
+
+openssl genrsa -out certs/ca.key 2048
+
+openssl req -new -x509 -days 3650 -key certs/ca.key -out certs/ca.crt -subj "/C=US/CN=landscape.local"
+```
```
YanisaHS: In general, code blocks should be able to be copied and pasted directly. Presumably these commands are sequential though and wouldn't be run at the same time?

If you're guiding the user to run them at the same time, then remove the blank lines in between.

If you don't want users to run them at the same time, then you should separate them into different code blocks with the instructions step by step (or at least have them in separate blocks with no text between, which is kind of fine for 2 blocks, but looks weird for more than 2 in a row)

Since yours has an example, I'd recommend putting it in a numbered list. So it's like

"...and private key with OpenSSL:

    1. To make the cert
    [command]
    1. To generate XYZ genrsa ...
    [command]
    1. To do XYZ openssl req thing ...
    [command]

After integrating the charms..."

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -92,18 +92,71 @@ See the [lego charm documentation](https://charmhub.io/lego) for DNS-01 challeng
...
+juju run manual-tls-certificates/0 provide-certificate \
+  certificate="$(base64 -w0 certs/client.crt)" \
+  ca-certificate="$(base64 -w0 certs/ca.crt)" \
+  certificate-signing-request="$(base64 -w0 certs/client.csr)"
+```
+
+Now, HAProxy is using the custom CA for TLS connections. You can verify HAProxy received the new TLS certificates using the `get-certificates` action on the HAProxy charm by providing the hostname of the configured Landscape Server root URL, for example:
+
+```sh
+juju run haproxy/0 get-certificate hostname=landscape.local --format=json | jq -r '.["haproxy/0"].["results"].certificate' > cert.pem
```
YanisaHS: Can you make the hostname more obviously a placeholder OR more obviously an example? To me `landscape.local` looks like it could be real / it isn't obvious the user should change it before running the command

YanisaHS (changes requested):
Mostly LGTM, although I have a few things. Up to you if you need a technical reviewer too

YanisaHS (approved):
🎀

PR #262, opened 2026-06-19
fix: clarify that custom CA certificates are supported

jansdhillon:
A customer asked whether they could use their own custom CA instead of the "generate a CSR" route. The docs were ambiguous and the example CA generation looked like a required step rather than just an example. This patch makes the intent unmistakable.

### DIFF
--- a/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
+++ b/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
@@ -109,7 +109,11 @@ juju integrate manual-tls-certificates:certificates haproxy:certificates
 juju integrate manual-tls-certificates:trust_certificate haproxy:receive-ca-certs
 ```
 
-Now, generate a CA certificate and private key or ensure they are available locally. For example, generate a CA certificate and private key with OpenSSL:
+Now, ensure you have a CA certificate and private key available locally. You can use any existing CA (such as your organisation's internal CA or corporate PKI) — the `manual-tls-certificates` charm works with any CA.
+
+```{tip}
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -109,7 +109,11 @@ juju integrate manual-tls-certificates:certificates haproxy:certificates
 juju integrate manual-tls-certificates:trust_certificate haproxy:receive-ca-certs
 ```
 
-Now, generate a CA certificate and private key or ensure they are available locally. For example, generate a CA certificate and private key with OpenSSL:
+Now, ensure you have a CA certificate and private key available locally. You can use any existing CA (such as your organisation's internal CA or corporate PKI) — the `manual-tls-certificates` charm works with any CA.
```
Perfect5th: —

(I'm cool with it)

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -109,7 +109,11 @@ juju integrate manual-tls-certificates:certificates haproxy:certificates
...
+
+```{tip}
+If you don't already have a CA, the following example shows how to generate a self-signed one with OpenSSL. **Skip this if you already have a CA certificate and key.**
```
Perfect5th: feels a little redundant to have both these sentences, but I guess the second one's **bold**?

Perfect5th (approved):

Perfect5th (approved):

PR #267, opened 2026-06-26
Migrate cve fix tutorial

YanisaHS:
Migrating the relevant information from: https://discourse.ubuntu.com/t/test-and-deploy-individual-cve-fixes-with-pro-client-and-landscape/27430


Perfect5th (approved):
LGTM but kind of sad that we don't have this as a "first-party" feature.

PR #270, opened 2026-07-07
docs: note that RabbitMQ 4.0+ is not supported

jansdhillon:
https://warthogs.atlassian.net/browse/LNDENG-4698

RabbitMQ 4.0 dropped the AMQP 0-8 protocol that Landscape depends on via txlongpoll. As a result, RabbitMQ 4.0 and later are not compatible with Landscape. Ubuntu 26.04 LTS (Resolute) ships RabbitMQ 4.x from its archives, which is why the Quickstart installation method is not supported on Ubuntu 26.04.

### DIFF
--- a/docs/how-to-guides/landscape-installation-and-set-up/configure-rabbitmq.md
+++ b/docs/how-to-guides/landscape-installation-and-set-up/configure-rabbitmq.md
@@ -8,7 +8,7 @@ myst:
 # How to configure RabbitMQ for Jammy 22.04 (or later)
 
 ```{note}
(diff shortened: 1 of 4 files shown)


docs/how-to-guides/landscape-installation-and-set-up/configure-rabbitmq.md
```diff
@@ -7,6 +7,10 @@ myst:
 (how-to-configure-rabbitmq)=
 # How to configure RabbitMQ for Jammy 22.04 (or later)
 
+```{note}
+Landscape requires RabbitMQ 3.x. RabbitMQ 4.0 dropped support for the AMQP 0-8 protocol that Landscape depends on, so RabbitMQ 4.0 and later are not supported. Ubuntu 26.04 LTS (Resolute) ships RabbitMQ 4.x, so Debian package installations of Landscape Server are not supported on Ubuntu 26.04. See {ref}`reference-known-issues` for details.
```
Perfect5th: "Debian package installations" doesn't seem quite accurate. You can do a manual installation using Debs if Rabbit is on a different machine (or in a container or something).

Perfect5th (changes requested):

Perfect5th (approved):

PR #271, opened 2026-07-07
feat(debarchive): include charm in juju deployment docs 

david-mclain:


david-mclain:
okay i messed something up hold on

PR #280, opened 2026-07-10
fix: clarify charm refresh doesn't upgrade

jmucc:
Feedback from support


YanisaHS (approved):

YanisaHS:
Ignore the pending check - it's due to me making some changes to our source project and webhooks.

PR #282, opened 2026-07-13
feat: add apache example to ubuntu installer config

HJK-X:
adds apache example

updates the note to discourage using quickstart. 

fix broken link

### DIFF
--- a/docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning.md
+++ b/docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning.md
@@ -10,11 +10,7 @@ myst:
 The Ubuntu installer (26.04 and later) can use Landscape to serve an autoinstall file. Your Landscape account must use OIDC authentication.
 
 ```{note}
-This feature is available from Landscape server `25.10` onwards.
-```
-
-```{note}
-This feature is available only on self-hosted deployments.
+This feature is available from Landscape server `25.10` onwards, and only on self-hosted deployments. It is not intended for {ref}`Quickstart <how-to-quickstart-installation>` deployments.
 ```
 
 ## Background information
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning.md
```diff
@@ -10,11 +10,7 @@ myst:
 The Ubuntu installer (26.04 and later) can use Landscape to serve an autoinstall file. Your Landscape account must use OIDC authentication.
 
 ```{note}
-This feature is available from Landscape server `25.10` onwards.
-```
-
-```{note}
-This feature is available only on self-hosted deployments.
+This feature is available from Landscape server `25.10` onwards, and only on self-hosted deployments. It is not intended for {ref}`Quickstart <how-to-quickstart-installation>` deployments.
```
YanisaHS:
```suggestion
This feature is available from Landscape server `25.10` onwards, and only on self-hosted deployments. {ref}`Quickstart <how-to-quickstart-installation>` deployments don't support using autoinstall.
```
If that's true. If it does technically support it but it's just not recommended, then leave as you originally had it.

docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning.md
```diff
@@ -94,7 +90,7 @@ This feature uses gRPC and requires an upstream proxy to perform HTTP/2 and TLS
 - The installer connects using HTTPS.
 - The `landscape-ubuntu-installer-attach` service listens on port 53354 by default.
 
-#### Example configuration (HAProxy)
+#### Example configuration
```
YanisaHS: I'd suggest having a clearer break, so it's more obvious that there are two different examples here.

#### Example configuration

**HAProxy**

[If you're using HAProxy, add the following..... (etc)]

**Apache**

[If you're using Apache, add the following...... (etc)]

YanisaHS: you can keep the H5 heading for Example Config (it didn't show in my comment, just looks bolded)

YanisaHS (approved):
LGTM although I've left a couple minor suggestions. Feel free to request me again if you want me to take another look, but otherwise it's fine not to require a re-review

Also though - assuming this is live now, you should re-target your PR to `main` before merging.

docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning.md
HJK-X: it does support it, just not recommended

YanisaHS (approved):
🎀

PR #284, opened 2026-07-14
add AI agent instructions and AI-use contribution guidance

YanisaHS:


AGENTS.md
```diff
@@ -0,0 +1,143 @@
...
+  through the web portal or the API.
+- The documentation covers different Landscape editions (SaaS, Managed, and
+  self-hosted), different interfaces (web portal and API), different installation
+  methods (for example Quickstart, manual, Juju/charm), and multiple product
+  releases. Applicability often differs between these.
+
+## Repository structure
+
+Documentation source lives under `docs/`. Content follows the
+[Diátaxis](https://diataxis.fr/) framework with some additions:
```
jansdhillon, line 28: Considering this is a link and a lot of the info about Diataxis is behind this, would it maybe make sense to tell agents to mention if this link is inaccessible?

jansdhillon (approved):
LGTM just a question

AGENTS.md
YanisaHS: We can change it later if it's not setting up stuff under Diataxis. I think that's more of an edge case where it can't access the external sources when the agent should have enough general context here and in neighboring docs to get the gist.

In general, if we have issues with that, I can add a new section that explains the 4 types. I think that'd be more helpful than using the link (which yea has a tonnn of info)

PR #287, opened 2026-07-20
Migrate docs tooling to Canonical Sphinx Stack 2.0

YanisaHS:
Notable changes:

**1. `conf.py`**
- Rewrote to mirror the 2.0 `conf.py` structure/ordering (easier future diffs)
- Kept Landscape-specific config: OGP/base URL, `source_edit_link`, static +
  template paths, `linkcheck_ignore`, heading anchors, cookie banner CSS/JS, `rst_prolog` roles, and `substitutions.yaml`
- Simplified the footer copyright to the year only, so it matches other Canonical docs
- Removed the old in-`conf.py` redirects dict and the `sphinx_reredirects` extension (redirects now live in `redirects.txt` via rediraffe)

**2. `requirements.txt`**
- Minimized/pinned to the 2.0 set; dropped unnecessary extensions
- Kept `sphinx-roles` (needed for the `{spellexception}` role used in our docs)

**3. Makefile**
- `docs/Makefile` now matches v2.0 (still with root `Makefile` forwards to `docs/`)

**4. added `_dev/` and other config files**
- Renamed `.sphinx/` to `_dev/` (2.0 layout)
- Kept a Landscape-specific `_dev/.pymarkdown.json` (has a few customizations)
- Removed the orphaned `reuse/links.txt`

**5. Redirects: migrate to rediraffe (`redirects.txt`)**
- Moved the legacy redirects into `redirects.txt` so we don't have two different redirect libraries

**6. Footer**
- Restored `_templates/footer.html` (2.0 built-in footer plus the "Manage your tracker settings" link) so the footer matches other Canonical docs

**7. updated `AGENTS.md` for the new paths**

**8. Test that everything works as expected**
- All `make` commands run, including `make pdf`
- **Migrated redirects**: I spot-checked locally
- **Sitemap**: generated; note it's renamed to `doc-sitemap.xml` (visit
  `/doc-sitemap.xml` instead of just `/sitemap.xml`)
- Edit button / feedback button still point to the right place
- Verified includes/reused content still works


YanisaHS:
Ignore the Check for removed URLs failure - some internal items moved around but users wouldn't have those as direct links

YanisaHS:
Will be reviewed by someone on the Docs team

AnneCYH:
Hey @YanisaHS , the changes seem to be in order :D 
I didn't check any project-specific conditions but the sphinx-stack 2.0 related updates are all good!

Approved!

PR #291, opened 2026-07-22
add TLS version and cipher hardening for reverse proxy and Postfix

YanisaHS:


YanisaHS:
Approved by Support (via MM)

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -23,19 +23,90 @@ Port 80 is only needed for Landscape's repository mirroring features. If you don
...
+**HAProxy**
+
+For pre-26.04 {ref}`Juju <how-to-juju-ha-installation>` deployments, the classic `haproxy` charm exposes bind options for this:
+
+```bash
+juju config haproxy global_default_bind_options='no-sslv3 no-tlsv10 no-tlsv11'
+juju config haproxy global_default_bind_ciphers='HIGH:!aNULL:!NULL:!3DES:!DSS:!EXP:!PSK:!SRP:!MD5:!SHA:!CBC:!CAMELLIA'
+```
+
+This disables SSL 3.0 and TLS 1.0/1.1 and restricts the TLS 1.2 cipher suites, but doesn't affect TLS 1.3. Then restart HAProxy on each unit to apply the change:
```
jansdhillon, line 75: We can use

```sh
juju exec -a haproxy "<cmd>"
```

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -23,19 +23,90 @@ Port 80 is only needed for Landscape's repository mirroring features. If you don
...
+
+```bash
+juju ssh haproxy/0 "sudo systemctl restart haproxy.service"
+juju ssh haproxy/1 "sudo systemctl restart haproxy.service"
+...
+juju ssh haproxy/n "sudo systemctl restart haproxy.service"
+```
+
+```{note}
```
jansdhillon, line 81: Maybe we should also link to https://canonical.com/juju/docs/haproxy-charm/latest/explanation/security/ in case they add something like this

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -23,19 +23,90 @@ Port 80 is only needed for Landscape's repository mirroring features. If you don
...
+Landscape 26.04 LTS and later use a different HAProxy charm (the `2.8` channel) that doesn't expose these options, so these steps apply only to pre-26.04 deployments.
```
jansdhillon: 2.8 is the track, 2.8/stable is the channel

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -23,19 +23,90 @@ Port 80 is only needed for Landscape's repository mirroring features. If you don
 
 The other applications in your deployment only require enough network access to communicate with each other. Using the default configuration, applications listen on these ports for incoming traffic:
 
-* Landscape server: 6554, and 8080-9100, inclusive
+* Landscape server: 8080-9100 (inclusive) and 50052
```
jansdhillon: It's not actually inclusive though right? That would be a huge range. Also, the port ranges are basically (base port) + 1 for each worker, so I think we should just list the default base ports for all the services. I think it's probably worth breaking down which service has which port.

jansdhillon: Also this is missing the Ubuntu Installer Attach port, which is an optional service but maybe worth mentioning?

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -12,7 +12,7 @@ You have many options when hardening your Landscape deployment.
 
 ## Harden the network
 
-The only application in your Landscape server deployment that should be exposed to incoming external traffic is the reverse proxy, which is either HAProxy or Apache. The reverse proxies listen on ports 80 and 443 for HTTP and HTTPS traffic, respectively.
+The only application in your Landscape server deployment that should be exposed to incoming external traffic is the reverse proxy, which is either HAProxy or Apache. The reverse proxies listen on ports 80 and 443 for HTTP and HTTPS traffic, respectively. The reverse proxy also exposes port 6554 for gRPC over HTTP/2 traffic. For more details on networking, see {ref}`reference-internal-network-requirements`.
```
jansdhillon: Worth mentioning that the gRPC services of Landscape currently only support using cleartext gRPC (h2c) without TLS (h2) ... This is actually what I'm running into with the charm right now. The gRPC services (hostagent-messenger and ubuntu-installer-attach) doesn't support ALPN

jansdhillon (changes requested):

docs/how-to-guides/security/harden-your-deployment.md
YanisaHS: I'm not sure if it's inclusive - that was in the original guide. @Perfect5th can you weigh in?

Regarding the networking for different services, we do have that doc'd here: https://ubuntu.com/landscape/docs/reference/networking/internal-network-requirements/

Does that doc cover your concerns? I want to keep this hardening guide lighter weight about networking stuff (i.e., only mention the things that are necessary for the hardening tasks)

YanisaHS: Is there another port that needs to be mentioned in this guide? If yes, what's the port number? I thought it was also 50052 :)

I'm going to create a story (assigned to an engineer) to go through that [internal networking guide](https://ubuntu.com/landscape/docs/reference/networking/internal-network-requirements/), make sure it's up to date, etc., ideally scheduled in the next 1-2 pulses.

**Edit:** Nevermind for the first part - I found [in an existing doc](https://ubuntu.com/landscape/docs/how-to-guides/ubuntu-installer/configure-landscape-deployment-for-autoinstall-provisioning/) that the autoinstall port is 53354

jansdhillon: Well I guess the service is optional and this says "the default configuration" so I think it's fine for this guide.

docs/how-to-guides/security/harden-your-deployment.md
YanisaHS: @jansdhillon Can you give this one as an actual suggestion for me to accept? I'm not sure what you mean, like which command that should replace

docs/how-to-guides/security/harden-your-deployment.md
```diff
@@ -23,19 +23,90 @@ Port 80 is only needed for Landscape's repository mirroring features. If you don
...
+juju config haproxy global_default_bind_ciphers='HIGH:!aNULL:!NULL:!3DES:!DSS:!EXP:!PSK:!SRP:!MD5:!SHA:!CBC:!CAMELLIA'
+```
+
+This disables SSL 3.0 and TLS 1.0/1.1 and restricts the TLS 1.2 cipher suites, but doesn't affect TLS 1.3. Then restart HAProxy on each unit to apply the change:
+
+```bash
+juju ssh haproxy/0 "sudo systemctl restart haproxy.service"
+juju ssh haproxy/1 "sudo systemctl restart haproxy.service"
+...
+juju ssh haproxy/n "sudo systemctl restart haproxy.service"
```
jansdhillon:
```suggestion
juju exec --application haproxy "sudo systemctl restart haproxy.service"
```

docs/how-to-guides/security/harden-your-deployment.md
jansdhillon: Okay I added a new suggestion

docs/how-to-guides/security/harden-your-deployment.md
Perfect5th: Yes, it's inclusive, but it should be 8070-9100. Appserver's default starting port is 8070.

jansdhillon: Should we maybe clarify that this it doesn't actually listen on all those ports then?

jansdhillon (approved):

PR #302, opened 2026-08-14
add machinery for openapi-overlay

Perfect5th:
This change adds a Read the Docs pre-build step that applies an OpenAPI overlay to the debarchive API schema OpenAPI YAML spec.

It also updates the Makefile such that this is applied when building the docs locally.

To check this works, try `make serve` and go to http://127.0.0.1:8000/reference/api/debarchive and check that the example overlay "My local repository" is present in the built docs.

After this is merged, I'll create a follow-up PR that will add the real examples.


docs/_static/openapi-overlay.yaml
```diff
@@ -0,0 +1,20 @@
+# Hand-written request/response examples for the generated Debarchive API spec
+# (docs/_static/openapi.yaml). Applied at build time by docs/scripts/apply_overlay.py
+# to produce docs/_static/openapi-with-examples.yaml, which is what
+# docs/reference/api/debarchive.md renders.
+#
+# See openapi-examples-overlay.md for the full design, including why targets
```
HJK-X: Is this the openapi-examples-overlay.md from the internal docs repo?
not sure if we want to reference a private repo here

Perfect5th: Right… that's a good point. I'll strip out all references to the design doc then.

HJK-X (approved):
LGTM

PR #306, opened 2026-08-19
fix: inaccuracies in installation guides on 26.04

jansdhillon:
https://warthogs.atlassian.net/browse/LNDENG-4552

I also added some agents for testing in the future https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/create-custom-agents-for-cli

Every installation/setup guide under `docs/how-to-guides/landscape-installation-and-set-up/` was live-tested end-to-end against real targets (LXD containers, real Juju models via `../landscape-server-operator`, and a real FIPS-enabled Ubuntu Pro VM) rather than reviewed by reading alone. Bugs found this way were fixed; the diff was then minimized to keep only independently-confirmed changes.

### Bugs found & fixed

- **`manual-installation.md`**
  - `postgresql-contrib` doesn't exist on Resolute - package guidance made release-agnostic.
  - Config section is `[appserver]`, not `[landscape]`.
  - `/etc/default/landscape-server` no longer has `RUN_ALL`/`RUN_MSGSERVER` flags on current packages - only `RUN_CRON`/`UPGRADE_SCHEMA` remain.
- **`install-landscape-client.md`**
  - Registering against a self-signed-cert server requires `--ssl-public-key`; the guide didn't document this and registration silently failed without it.
- **`install-in-a-lxd-container.md`**
  - The documented cloud-init variable name didn't match the real template's expected variable (`LANDSCAPE_PPA`), so the profile-driven install silently used the wrong PPA.
- **`configure-landscape-client.md`**
  - The config-management (`RUN=1`) registration flow description didn't match observed behavior on current releases - corrected and simplified.
- **`juju-installation.md` / `juju-ha-installation.md`**
  - `landscape-scalable` on Charmhub has no `26.04/*` channel - clarified legacy vs. current-channel deploy paths.
  - HAProxy channel `2.8/edge` → `2.8/stable`.
  - **HAProxy hostname-routing bug**: the app is only reachable via the configured hostname, not the HAProxy unit's bare IP - this was silently wrong in both guides and is now documented.
- **`quickstart-installation.md`**
  - Corrected the Ubuntu Pro attach step wording after testing live with a real Pro token.
- **`disa-stig.md`**
  - PostgreSQL server cert needs a `DNS:localhost` SAN or localhost TLS connections fail verification.
  - `pg_hba.conf` example using the server's real IP is wrong for this all-in-one guide - corrected to loopback addressing.
  - RabbitMQ needs `NODENAME=rabbit@localhost` **and** a localhost TLS listener.
  - Removed stale `landscape-standalone-knowledge` package references.
  - Confirmed [LNDENG-4203](https://warthogs.atlassian.net/browse/LNDENG-4203) (PostgreSQL cert-auth setup bug) is fixed and backported to the `26.04` branch (canonical/landscape-server#1478), so no workaround note was needed - the straightforward cert-auth-from-the-start flow is correct as documented.
- **`install-landscape-in-an-air-gapped-or-offline-environment.md`**
  - Corrected steps found to not work as written when tested in a genuinely network-isolated container against a real offline apt mirror.
- **`install-on-fips-compliant-machines.md`**
  - Corrected steps found to not work as written when tested in a real FIPS-enabled Ubuntu Pro VM.

### Confirmed false alarms (no changes needed)

- `debarchive-repository-management.md` and `configure-outbox.md` were reviewed and tested; no inaccuracies found.
- `configure-postfix.md`, `configure-rabbitmq.md`, `create-saas-account.md` reviewed; no inaccuracies found.

### Validation

- `make html` run clean after every change.
- All fixes reproduced live before being applied - diff was subsequently minimized to remove any change not independently confirmed necessary (see commit history on this branch).

### Tooling added

- `.github/agents/tester-*.agent.md`: one custom tester agent definition per guide, capturing the live-testing methodology (real LXD containers/VMs, real Juju models, real FIPS/Pro tokens) used for this audit, so future re-tests of these guides can be run consistently.

[LNDENG-4203]: https://warthogs.atlassian.net/browse/LNDENG-4203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client.md
```diff
@@ -175,7 +175,11 @@ sudo chmod 777 <CUSTOM_TEMPDIR>
 
 ## Landscape clients with configuration management tools
 
-If you want to manage `landscape-client` through a configuration management tool such as Puppet or Ansible, you can avoid getting duplicate computers by writing the `/etc/landscape/client.conf` and `/etc/default/landscape-client` files, and then restarting the `landscape-client` service.
+If you want to manage `landscape-client` through a configuration management tool such as Puppet or Ansible, you can avoid getting duplicate computers by writing the `/etc/landscape/client.conf` file yourself and then running `landscape-config --silent` to apply it.
+
+```{note}
+The `landscape-client` systemd service only starts once a registration request has been sent. Just writing `client.conf` and restarting the service isn't enough — running `landscape-config --silent` is what actually triggers registration and lets the service start.
```
YanisaHS: I don't think this note is needed - you edited the paragraph above already to add that you need to run `landscape-config` to apply the changes. And in the [section on editing client.conf, it also says landscape-config is needed to apply changes](https://ubuntu.com/landscape/docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client/#update-the-client-conf-file) (although actually there it might be better to say "run" landscape-config instead of "use", but anyway, the information is covered elsewhere)

docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client.md
```diff
@@ -192,13 +196,13 @@ include_manager_plugins = ScriptExecution
...
-In `/etc/default/landscape-client`:
+Then, apply the configuration and trigger registration with:
 
 ```bash
-RUN=1
+sudo landscape-config --silent
 ```
 
-The advantage over calling `landscape-config` is that this will request a registration only if the client is not already registered against `landscape-server`. Be aware that some configuration options (namely `computer_title`, `tags`, `access_group`) are only sent to `landscape-server` on registration.
+The advantage over calling `landscape-config` interactively is that `--silent` reads all settings from the existing `client.conf` and requests a registration only if the client is not already registered against `landscape-server`. Be aware that some configuration options (namely `computer_title`, `tags`, `access_group`) are only sent to `landscape-server` on registration.
```
YanisaHS: I think this is good information, but can you move it up to the [Use landscape-config](https://ubuntu.com/landscape/docs/how-to-guides/landscape-installation-and-set-up/configure-landscape-client/#use-landscape-config) section instead?

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -181,7 +181,7 @@ For PostgreSQL, you'll need three certificates and their corresponding keys:
     sudo chmod 400 /etc/landscape/postgres_client_superuser.key
     ```
 
-- Server authentication certificate. The SAN must contain the DNS or IP address of the database server.
+- Server authentication certificate. The SAN must contain the DNS or IP address of the database server. For the all-in-one deployment in this guide, include `DNS:localhost` in the SAN because Landscape connects to PostgreSQL using `host = localhost`.
```
YanisaHS: Can you reword "the all-in-one deployment in this guide"? I'd provide a suggestion, but I'm not sure what you mean by the "all-in-one" deployment?

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -233,16 +233,13 @@ Use the following steps to harden the PostgreSQL service.
 
 PostgreSQL must be configured to allow the Landscape application server to access the database server. Landscape uses several users for access, so all users must be added.
 
-Edit the file `/etc/postgresql/14/main/pg_hba.conf` and add:
+For the all-in-one deployment in this guide, Landscape connects to PostgreSQL using `host = localhost`, so use loopback entries instead of the server's external IP address. Edit the file `/etc/postgresql/14/main/pg_hba.conf` and add:
```
YanisaHS: Same here, can you use whatever wording you came up with before to replace "the all-in-one deployment"?

docs/how-to-guides/landscape-installation-and-set-up/disa-stig.md
```diff
@@ -1108,9 +1103,11 @@ stores = main account-1 resource-1
...
 
 [schema]
-# note that you must have at least two certificates for db connections:
-# one for landscape_superuser
-# and one for the regular landscape user
+# [schema] ssl settings apply only to the superuser (landscape_superuser)
+# connection. They are independent of the [stores] ssl settings, which
+# apply to the regular landscape user connection.
+# PostgreSQL certificate authentication requires the certificate CN to match
+# the connecting username, so each role must have its own client certificate.
```
YanisaHS: Is there a reason this information is i nthe file rather than in the text? If not, you should move it to be part of the regular doc instead of a comment in the config file

docs/how-to-guides/landscape-installation-and-set-up/install-landscape-client.md
```diff
@@ -184,12 +184,16 @@ Download the server certificate:
...
+Then register the client, passing the certificate you just downloaded with `--ssl-public-key` so the client can verify the server:
 
 ```bash
-sudo landscape-config --computer-title "My client" --account-name standalone --url https://{LANDSCAPE_FQDN}/message-system --ping-url http://{LANDSCAPE_FQDN}/ping
+sudo landscape-config --computer-title "My client" --account-name standalone --url https://{LANDSCAPE_FQDN}/message-system --ping-url http://{LANDSCAPE_FQDN}/ping --ssl-public-key /etc/landscape/server.pem
 ```
 
-If you used a custom CA, you'll need to pass the `--ssl-public-key` parameter pointing to the CA file so that the client can recognize the issuer of the server certificate.
+```{note}
+`--ssl-public-key` is only needed when the client doesn't already trust the server's certificate — for example, this self-signed certificate, or one signed by a custom/private CA. A certificate from a publicly trusted CA (for example, via Let's Encrypt/`lego`) is already trusted by the client's system CA store, so the flag isn't needed in that case.
```
YanisaHS:
```suggestion
`--ssl-public-key` is only needed when the client doesn't already trust the server's certificate. For example, this self-signed certificate, or one signed by a custom/private CA. A certificate from a publicly trusted CA (for example, via Let's Encrypt/`lego`) is already trusted by the client's system CA store, so the flag isn't needed in that case.
```
No reason other than AI has made me 🔥 **hate** 🔥  those dashes

docs/how-to-guides/landscape-installation-and-set-up/juju-installation.md
```diff
@@ -30,62 +30,28 @@ If you have an Ubuntu Pro subscription, attach your Pro token to each machine th
...
-Starting with the **26.04 beta version** of the Landscape Server charm, the deployment architecture has changed significantly. The charm now integrates directly with the external HAProxy charm (`2.8/edge`) using the `haproxy-route` interface and no longer uses the legacy `reverseproxy` interface. If you have an existing deployment using the older approach, see {ref}`how-to-migrate-to-26-04-charm` for migration instructions.
-```
-
-### Deployment approaches
-
-There are two deployment approaches depending on which version of the Landscape Server charm you're using:
+Starting with the **26.04 beta version** of the `landscape-server` charm, the deployment architecture changes to PostgreSQL 14+ over the `database` relation (backed by the `postgresql_client` charm interface), HAProxy 2.8 over `haproxy-route`, and TLS via `tls-certificates`.
 
-#### Pre-26.04 deployment
+The Charmhub `landscape-scalable` bundle is deprecated and does not have a `26.04/*` track. For 26.04+ deployments, follow {ref}`how-to-juju-ha-installation`.
```
YanisaHS:
```suggestion
The Charmhub `landscape-scalable` bundle doesn't have a `26.04/*` track. For 26.04+ deployments, follow {ref}`how-to-juju-ha-installation`.
```
Because presumably it's not actually "deprecated", right? It just doesn't have a 26.04 track? (but we still technically have an older track of the bundle we support)

If `landscape-scalable` bundles are truly deprecated, no one uses and we don't support those who use them, we should have a deprecation note here as well https://ubuntu.com/landscape/docs/how-to-guides/landscape-installation-and-set-up/juju-installation/#landscape-scalable-bundle, and on Charmhub, etc.

**Edit:** Actually now looking through your whole page edit more closely, I see you've revamped it to cover that stuff. If we're calling it deprecated, we should have a date though, maybe something like "The bundle was deprecated in 26.04, there's no 26.04 track", to provide clarity on when it was considered deprecated

docs/how-to-guides/landscape-installation-and-set-up/juju-installation.md
```diff
@@ -102,8 +68,10 @@ Once the deployment has finished, Landscape Server is accessible in different wa
 
 **26.04 beta+ deployment:**
 
-  - Access via the HAProxy unit IP address or your configured `root_url`
-  - HAProxy handles load balancing across all Landscape Server units
+  - HAProxy routes traffic based on the `hostname` set in the `haproxy-route` relation, **not** by port alone, so you must connect using that hostname (for example, with `--resolve` or a DNS entry pointing it at the HAProxy unit's IP) — connecting directly to the HAProxy unit's own IP address (with no matching `Host` header) hits HAProxy's default page instead of Landscape.
+  - If you set `root_url`, that hostname is what you must connect with.
+  - If you leave `root_url` unset, the `landscape-server` charm falls back to using the leader unit's IP address as the routing hostname, so you'd need to connect using *that* IP as the `Host` header/SNI value (not the HAProxy unit's IP) — in practice, setting `root_url` to a real hostname is much simpler for testing.
```
YanisaHS:
```suggestion
  - HAProxy routes traffic based on the `hostname` set in the `haproxy-route` relation, **not** by port alone, so you must connect using that hostname (for example, with `--resolve` or a DNS entry pointing it at the HAProxy unit's IP). Connecting directly to the HAProxy unit's own IP address (with no matching `Host` header) hits HAProxy's default page instead of Landscape.
  - If you set `root_url`, that hostname is what you must connect with.
  - If you leave `root_url` unset, the `landscape-server` charm falls back to using the leader unit's IP address as the routing hostname, so you'd need to connect using *that* IP as the `Host` header/SNI value (not the HAProxy unit's IP). In practice, setting `root_url` to a real hostname is much simpler for testing.
```

docs/how-to-guides/landscape-installation-and-set-up/manual-installation.md
```diff
@@ -55,7 +55,11 @@ sudo apt install postgresql postgresql-16-debversion postgresql-plpython3-16 pos
 For an Ubuntu 26.04 ("resolute") database server:
 
 ```bash
-sudo apt install postgresql postgresql-18-debversion postgresql-plpython3-18 postgresql-contrib
+sudo apt install postgresql postgresql-18-debversion postgresql-plpython3-18
+```
+
+```{note}
+Starting with Resolute, PostgreSQL no longer has a separate `postgresql-contrib` package. Its extensions are now bundled directly into the `postgresql-<version>` package, so no extra package is required.
```
YanisaHS:
```suggestion
Starting with Ubuntu 26.04 LTS, PostgreSQL no longer has a separate `postgresql-contrib` package. Its extensions are now bundled directly into the `postgresql-<version>` package, so no extra package is required.
```
You could also add "Resolute Raccoon" if you want. But also this is a bit unclear to me, you're saying the PostgreSQL version that's shipped with Ubuntu doesn't have this package? It's unclear to me because presumably a user could use whatever PostgreSQL version they want, including an older one, so the limitation isn't inherently the Ubuntu OS?

YanisaHS (changes requested):

docs/how-to-guides/landscape-installation-and-set-up/install-on-fips-compliant-machines.md
YanisaHS, line 1: I just noticed that this doc actually says they can use Landscape 26.04 for FIPS, is that right? I know it has to be on Jammy now, but your airgap guide says "Landscape 26.04 LTS depends on two snap services, landscape-outbox and landscape-debarchive. Both snaps currently use the core24 base" - isn't only core22 FIPS?

docs/how-to-guides/landscape-installation-and-set-up/install-landscape-in-an-air-gapped-or-offline-environment.md
```diff
@@ -45,9 +45,11 @@ All of the necessary packages for Landscape Server should now be downloaded to t
 
 #### (Landscape 26.04 only) Download the Landscape Server snap services
 
-Landscape 26.04 LTS depends on two snap services, `landscape-outbox` and `landscape-debarchive`, which are downloaded separately.
+Landscape 26.04 LTS depends on two snap services, `landscape-outbox` and `landscape-debarchive`. Both snaps currently use the `core24` base and require the `snapd` system snap to be available on the offline machine, so download all four snaps separately:
```
YanisaHS: Relevant to this comment https://github.com/canonical/landscape-documentation/pull/306/changes#r3823171904

YanisaHS:
I only scanned through the agents tests part though, LMK if you want me to look at that closer.

**Edit:** Thinking about it again, should the test agents just be in another PR so someone (maybe me, but presumably an engineer) can look at them more closely?

docs/how-to-guides/landscape-installation-and-set-up/install-landscape-in-an-air-gapped-or-offline-environment.md
YanisaHS: @jansdhillon We discussed this in standup - can you check in with the squads on these snaps what the status is? I believe these snaps use core22 for now

docs/how-to-guides/landscape-installation-and-set-up/install-on-fips-compliant-machines.md
jansdhillon: Confirmed with the engineering source (canonical/landscape-go's outbox/snap/snapcraft.yaml and debarchive/snap/snapcraft.yaml, plus `snap info landscape-outbox` on an installed instance): both landscape-outbox and landscape-debarchive declare `base: core22`, not core24. Fixed the air-gapped guide accordingly (40a3e72). So this doc's FIPS applicability claim is correct - these snaps are core22-based, consistent with the FIPS guide's core22 requirement.

jansdhillon:
> I only scanned through the agents tests part though, LMK if you want me to look at that closer.
> 
> **Edit:** Thinking about it again, should the test agents just be in another PR so someone (maybe me, but presumably an engineer) can look at them more closely?

Sure

docs/how-to-guides/landscape-installation-and-set-up/install-landscape-in-an-air-gapped-or-offline-environment.md
jansdhillon: Yep, I've check and they both use core22

YanisaHS (approved):

PR #317, opened 2026-08-27
docs: add explanation for package change plans

abehnia:


docs/explanation/features/package-change-plans.md
```diff
@@ -0,0 +1,52 @@
...
+myst:
+  html_meta:
+    description: "Understand how Landscape package change plans use a snapshot of package data on Landscape Server, and how that can differ from the actual state on your instances."
+---
+
+(explanation-package-change-plan-snapshot)=
+# Package change plans
+
+Available in Landscape 26.10 and later.
+
```
YanisaHS, line 22: Can you briefly describe what a package change plan is? The explanation starts too close to the implementation details, with how it's built and where the data comes from.

docs/explanation/features/package-change-plans.md
```diff
@@ -0,0 +1,52 @@
...
+## What this means when you create a plan
+
+- The package versions, installed and held states, and available upgrades shown
+  in a plan come from the snapshot, not from the instance itself.
+- Instances that cannot perform an action according to the snapshot
+  are listed as exclusions. Those exclusions are also based on the
+  snapshot.
+- The snapshot is taken when the plan is created and is not refreshed when the
+  plan is executed. Changes made on an instance in the meantime, locally or
+  through another Landscape activity, are not reflected in the plan.
```
YanisaHS, line 43:
```suggestion
- The plan records the package state available to Landscape Server when the plan is created. This state is not refreshed when the plan is executed. Changes made to an instance after the plan is created, either locally or through another Landscape activity, are not reflected in the plan.
```
Something like this instead?

The main issue to me though was saying "the snapshot is taken" feels contradictory to the distinction you were making above, that it's a snapshot reflecting the reported state, not necessarily a snapshot taken at that moment

docs/explanation/features/index.md
```diff
@@ -32,6 +32,7 @@ Landscape enables you to control software across your instances, including mirro
 Repository mirroring (pre-26.04)<repository-mirroring>
 Repository mirroring (26.04+)<repository-mirroring-2604>
 Package reporting <package-reporting>
+Package Change Plans <package-change-plans>
```
YanisaHS:
```suggestion
Package change plans <package-change-plans>
```

docs/explanation/features/package-change-plans.md
```diff
@@ -0,0 +1,52 @@
...
+repositories Landscape knows about.
+
+Landscape does not query the instances while you build a plan. The plan is a
+snapshot: it reflects the last reported state, which may be minutes or hours
+old, or older for instances that have been offline.
+
+## What this means when you create a plan
+
+- The package versions, installed and held states, and available upgrades shown
+  in a plan come from the snapshot, not from the instance itself.
```
YanisaHS:
```suggestion
- The package versions, installed and held states, and available upgrades shown
  in a plan come from Landscape Server's stored package state, not a live query to the instance.
```
Is that more precise? Because the snapshot _does_ ultimately come from the instance, right?

docs/explanation/features/package-change-plans.md
```diff
@@ -0,0 +1,52 @@
...
+- An operation can fail or do nothing if the instance's state changed after the
+  plan was created.
+- The package manager resolves dependencies on the instance, so it may install,
+  upgrade, or remove packages that the plan does not list.
+
+Check the resulting activities for the actual per-instance outcome.
+
+## Recommendations
+
+- Keep instances checking in regularly so the snapshot stays current.
```
YanisaHS:
```suggestion
- Keep instances checking in regularly so Landscape has recent package state when creating the plan. The available package information can still depend on when the instance last refreshed its package cache.
```
Right? I could go either way on this one though. I felt like leaving that recommendation without additional details implies that if their instances check in regularly, then their inherently current, although if I understood package reporting correctly, the cache can still be a source of data being out-of-date

YanisaHS (changes requested):

srunde3 (commented):
Not sure if you'd already seen the draft PR I had - I include an explanation there. Feel free to pull from it as necessary if you haven't already: https://github.com/canonical/landscape-documentation/pull/256/changes#diff-376346954bcd4efbfce8f385f3d7689d7e44a182d79cac8d0a0d16adaee41546

docs/explanation/features/package-change-plans.md
abehnia: My goal was to explain the parts which were "interesting" or "ambiguous" to the user. I feel describing the plan itself would be redundant since it's just performing an action install/remove/etc on computers.

docs/explanation/features/package-change-plans.md
abehnia: Yes, and no. There are two levels: 1. The information that Server has is not exactly the same as what the instances see. 2. When the change plan is created, it never recomputes the plan so in-effect, it's a snapshot.

docs/explanation/features/package-change-plans.md
abehnia: yeah it'll never match up with the instances' state. We can potentially change the current wording to be more approximate.

docs/explanation/features/package-change-plans.md
abehnia: The original wording is accurate, if changes happen to the stored package state on Landscape Server, the plan is not going to change, hence the snapshot wording.

docs/explanation/features/package-change-plans.md
```diff
@@ -0,0 +1,64 @@
...
+- An operation can fail or do nothing if the instance's state changed after the
+  plan was created.
+- The package manager resolves dependencies on the instance, so it may install,
+  upgrade, or remove packages that the plan does not list.
+
+Check the resulting activities for the actual per-instance outcome.
+
+## Recommendations
+
+- Keep instances checking in regularly so the snapshot stays reasonbly up to date.
```
YanisaHS:
```suggestion
- Keep instances checking in regularly so the snapshot stays reasonably up to date.
```

YanisaHS (approved):
LGTM (just the small typo), but you should wait until after the point release to merge.

srunde3 (approved):
What you have is accurate, LGTM.

In my explanation draft I included some information about what could result in a computer being "excluded" for certain actions, especially `change_version`. I think that could just as easily go into the API reference, if it's necessary at all.

PR #336, opened 2026-09-01
feat: document deprecation of packages endpoints

wck0:
See https://github.com/canonical/landscape-server/pull/1614


YanisaHS (approved):
LGTM, presumably the functionality is being replaced by another endpoint? If that endpoint exists already, you should add another line guiding users what endpoint to use instead though

wck0:
> LGTM, presumably the functionality is being replaced by another endpoint? If that endpoint exists already, you should add another line guiding users what endpoint to use instead though

The replacements are a work in progress. When they are created, adding a link from the deprecated endpoints to the new ones will be part of that docs PR.

PR #340, opened 2026-09-04
Crosslink to juju upgrade steps

### LINKED ISSUE #339: Cross link back to upgrade-landscape page to differentiate charm refresh with package-upgrade
In Step 4 of "[How to migrate to the 26.04 charm](https://ubuntu.com/landscape/docs/how-to-guides/upgrade/migrate-to-26-04-charm/#step-4-refresh-the-charm)" says the PPA must be updated manually on each unit, but there is no link to the detailed procedure. The detailed procedure is at https://ubuntu.com/landscape/docs/how-to-guides/upgrade/upgrade-landscape/ 

We should have a link back to the "upgrade-landscape" page to clarify the process.

YanisaHS:
Closes #339

### DIFF
--- a/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
+++ b/docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
@@ -190,7 +190,7 @@ juju refresh landscape-server --channel 26.04/stable
 ```{note}
 `juju refresh` updates the charm revision, but it does not upgrade the installed `landscape-server` deb packages on existing units.
 
-The landscape-server charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the landscape-server units.
+The landscape-server charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the landscape-server units. For the PPA and package upgrade steps, see {ref}`how-to-heading-upgrade-juju`. You've already refreshed the charm, so skip the `juju refresh landscape-server` step at the start of that procedure.
 ```
 
 Wait for the refresh to complete and the services to restart:
--- a/.github/workflows/sync-main-to-develop.yml
+++ b/.github/workflows/sync-main-to-develop.yml
@@ -18,6 +18,7 @@ jobs:
 
     env:
       GH_TOKEN: ${{ secrets.LANDSCAPE_SYNC_TOKEN }}
+      SYNC_BRANCH: sync-main-into-develop
 
     steps:
       - name: Check whether develop is already up to date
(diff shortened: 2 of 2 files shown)


docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -190,7 +190,7 @@ juju refresh landscape-server --channel 26.04/stable
 ```{note}
 `juju refresh` updates the charm revision, but it does not upgrade the installed `landscape-server` deb packages on existing units.
 
-The landscape-server charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the landscape-server units.
+The landscape-server charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the landscape-server units. For the PPA and package upgrade steps, see {ref}`how-to-heading-upgrade-juju`. You've already refreshed the charm, so skip the `juju refresh landscape-server` step at the start of that procedure.
```
jansdhillon, line 193: Actually now that I'm reading this I think this is no longer true for the 26.04+ charm: https://github.com/canonical/landscape-server-operator/pull/77

So I guess this is also wrong: https://ubuntu.com/landscape/docs/how-to-guides/upgrade/migrate-to-26-04-charm/#step-4-refresh-the-charm

Or rather, they could basically do what I put in the manual testing of that PR (just change the charm config, run upgrade action on each unit)

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
```diff
@@ -190,7 +190,7 @@ juju refresh landscape-server --channel 26.04/stable
 ```{note}
 `juju refresh` updates the charm revision, but it does not upgrade the installed `landscape-server` deb packages on existing units.
 
-The landscape-server charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the landscape-server units.
+Set `landscape_ppa` to `ppa:landscape/self-hosted-26.04` in the charm configuration, then follow the package upgrade steps in {ref}`how-to-heading-upgrade-juju`. The 26.04 charm's `upgrade` action adds the configured PPA before it upgrades the packages, so you don't need to add it manually on each unit. You've already refreshed the charm, so skip the `juju refresh landscape-server` step at the start of that procedure.
```
jansdhillon, line 193: Also sorry now that I'm thinking about it AGAIN I don't think it says anywhere that you can actually provide multiple PPAs when they're comma-separated via the charm config: https://github.com/canonical/landscape-server-operator/pull/110... Maybe just a ```{tip}``` would be nice here?

docs/how-to-guides/upgrade/upgrade-landscape.md
```diff
@@ -96,13 +96,15 @@ To upgrade a basic Juju deployment:
...
-    juju ssh landscape-server/0 "sudo add-apt-repository ppa:landscape/self-hosted-24.04 -y"
     ```
 
     If you’re upgrading to a different version of Landscape, use the appropriate PPA.
 
-    **Note**: At the moment, the `landscape-server` charm only adds the PPA source during installation, so you will need to manually update this PPA on each of the `landscape-server` units. When the charm is updated, this will no longer be necessary.
+    **Note**: From the 26.04 version of the `landscape-server` charm onwards, the `upgrade` action adds the configured PPA before it upgrades the packages, so no further action is needed. On earlier versions of the charm, the PPA source is only added during installation, so you must also add it manually on each of the `landscape-server` units:
+
+    ```bash
+    juju ssh landscape-server/0 "sudo add-apt-repository ppa:landscape/self-hosted-24.04 -y"
```
jansdhillon:
```suggestion
    juju exec --application landscape-server "sudo add-apt-repository ppa:landscape/self-hosted-24.04 -y"
```

This just does it on all the units at once (plus `juju ssh` is sometimes restricted while `juju exec` basically always works)

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
YanisaHS: @jansdhillon can you give me an example of what that would look like? (I'm not sure why you would need to provide multiple Landscape PPAs)

jansdhillon: Something like this:

```suggestion
```{tip}
The 26.04 version of the `landscape-server` charm can add multiple PPAs during install and when running the `upgrade` action when provided to the `landscape_ppa` charm config option in a comma-separated format, for example:

    juju config landscape-server landscape_ppa="ppa:landscape/self-hosted-26.04,ppa:canonical-python-maintainers/python-backports"

```

But yeah you're right, I'm not sure anyone would ever run into this since we advise to run Landscape on it's own machine, so unless they were using a custom version of Landscape from their PPA... probably not even worth it honestly

jansdhillon (approved):

docs/how-to-guides/upgrade/migrate-to-26-04-charm.md
YanisaHS: ok! I get it - I was confused how they could add multiple _Landscape_ PPAs. But yea since you're iffy on it anyway, I don't want to give this guidance because it doesn't semantically fit with the variable name

PR #342, opened 2026-09-04
fix(ci): open the sync PR from a branch and merge it when clean

YanisaHS:
The sync PR used main as its head branch, so conflicts could only be resolved by committing develop into main, which would pull unreleased docs onto main. Open the PR from a disposable sync-main-into-develop branch instead, so conflicts can be resolved on the PR itself.

Also merge the PR outright when GitHub reports it as mergeable, falling back to auto-merge only when required checks or reviews block it.

Tested that it works [here](https://github.com/canonical/landscape-documentation/actions/runs/33908206918/job/101138114151)

### DIFF
--- a/.github/workflows/sync-main-to-develop.yml
+++ b/.github/workflows/sync-main-to-develop.yml
@@ -5,6 +5,17 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      source_branch:
+        description: "Branch to sync from. Override only to test the workflow."
+        required: false
+        default: main
+        type: string
+      target_branch:
+        description: "Branch to sync into. Override only to test the workflow."
+        required: false
+        default: develop
+        type: string
 
 concurrency:
   group: sync-main-to-develop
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-main-to-develop.yml
```diff
@@ -5,6 +5,17 @@ on:
...
+      source_branch:
+        description: "Branch to sync from. Override only to test the workflow."
+        required: false
+        default: main
+        type: string
+      target_branch:
+        description: "Branch to sync into. Override only to test the workflow."
+        required: false
+        default: develop
+        type: string
```
jmucc: Did you want to keep the override functionality? I don't think it has to be removed personally but wanted to make sure.

jmucc (approved):
LGTM

.github/workflows/sync-main-to-develop.yml
YanisaHS: I meant to remove this - I was using it to test if the action work

Also hardcoded the branches to main and develop

jmucc (approved):
LGTM

PR #347, opened 2026-09-09
Add admin limit information

gesquivelgaghi:
- Add `GET /max-people-count` endpoint
- Add a line in the UI about the admin limit

### DIFF
--- a/docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
+++ b/docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
@@ -30,7 +30,7 @@ Users who receive an invitation will see an HTML link in the email. Clicking thi
 
 The first person to click on the link and submit information becomes an administrator, even if it’s not the person with the name and email address to which you sent the invitation. Take care to keep track of the status of administrator invitations.
 
-There is a default limit of 10 administrators for an account. To increase that limit, reach out to support.
+There is a default limit of 5 administrators for an account. To increase that limit, reach out to support.
 
 ## Remove administrators
 
--- a/docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
+++ b/docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
@@ -30,6 +30,8 @@ Users who receive an invitation will see an HTML link in the email. Clicking thi
 
 The first person to click on the link and submit information becomes an administrator, even if it’s not the person with the name and email address to which you sent the invitation. Take care to keep track of the status of administrator invitations.
 
+There is a default limit of 10 administrators for an account. To increase that limit, reach out to support.
+
 ## Remove administrators
 
 To remove an administrator from Landscape:
--- a/docs/reference/api/rest-api-endpoints/administrators.md
+++ b/docs/reference/api/rest-api-endpoints/administrators.md
@@ -70,6 +70,6 @@ Example response:
 
 ```json
 {
-  "max_people_count": 10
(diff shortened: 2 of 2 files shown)


docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
```diff
@@ -30,6 +30,8 @@ Users who receive an invitation will see an HTML link in the email. Clicking thi
 
 The first person to click on the link and submit information becomes an administrator, even if it’s not the person with the name and email address to which you sent the invitation. Take care to keep track of the status of administrator invitations.
 
+There is a default limit of 5 administrators for an account. To increase that limit, reach out to support.
```
gesquivelgaghi, line 33: Looked a bit more into it and Copilot said: `Note that standalone (self-hosted/LDS) accounts use a different default: DEFAULT_STANDALONE_ACCOUNT_MAX_PEOPLE_COUNT = 1000.`

@YanisaHS How do we address these differences within the docs?

YanisaHS: @gesquivelgaghi 

```suggestion
Landscape SaaS accounts have a default limit of 10 administrators. Pending administrator invitations count toward this limit. To request a higher limit, contact Canonical Support.
```

Reasons for my changes:
- Specified SaaS
- Added the pending part because [I found this in your UI PR](https://github.com/canonical/landscape-ui/pull/772#pullrequestreview-5157280153), it's useful info to include
- Changed it to **10** (not 5) admins - this is what you said the limit was in MM? (of course change it back if I'm wrong though :))
- Changed the wording to be "request" an increase, rather than make it seem like they just increase it regardless (not sure if customer-facing teams have prereqs for allowing more admins, or if they will introduce that in the future, just making it more clear that it's a request)
- Made it more clear what "support" is. It'd also be fine to use a capital S instead, like "contact Support"
- "contact" is clearer than "reach out", particularly for lower English level users (this is just a nit I added since I was in here anyway, not a big deal)

> How do we address these differences within the docs?

The SaaS vs self-hosted question is implicitly addressed just by calling out SaaS. Since the "limit" for self-hosted is 1000, I wouldn't even call this out in the docs since that's not needed information for a user - it's unlikely someone would need 1000+ admins. Presumably self-hosted people could change it anyway if needed. BUT I'm not against adding that information if it's useful, then I'd write it like 

"Landscape SaaS accounts have a default limit of 10 administrators; for self-hosted accounts, the default limit is set to 1000. Pending administrator invitations count toward this limit. 

To request a higher limit for SaaS accounts, contact Canonical Support."

docs/reference/api/rest-api-endpoints/administrators.md
```diff
@@ -47,3 +47,29 @@ Example response:
...
+
+```bash
+curl -X GET https://landscape.canonical.com/api/v2/max-people-count -H "Authorization: Bearer $JWT"
+```
+
+Example response:
+
+```json
+{
+  "max_people_count": 5
```
YanisaHS, line 73: I'd match this with the SaaS default, so if that's 10, make it 10 here.

YanisaHS (changes requested):

docs/how-to-guides/web-portal/web-portal-24-04-or-later/manage-administrators-and-roles.md
gesquivelgaghi: Yeah, turns out its actually 5 default, 10 is increased.

YanisaHS (approved):
LGTM then! you can ignore the linkcheck - I'm opening a PR in a bit to fix that

PR #351, opened 2026-09-15
fix(disa-stig): use -- instead of emdash for client SSL public key

jansdhillon:
https://warthogs.atlassian.net/browse/LNDENG-5061


Perfect5th (approved):
lol

PR #354, opened 2026-09-15
fix(manual-installation): specify secret_token section per Landscape release

YanisaHS:
This is a request from Support.


jmucc (approved):
Thanks Yanisa! I also noticed the linkcheck thing with the terraform docs, nice fix there as well!

PR #356, opened 2026-09-18
fix: add workaround for migration to debarchive for deployments prior to 25.10

jmucc:
## Jira Ticket
https://warthogs.atlassian.net/browse/LNDENG-5172

## landscape-go PR
https://github.com/canonical/landscape-go/pull/403


david-mclain (approved):
Approving though small question, if they use the fix provided doesn't it kind of resolve this issue? Granted the fix you made probably won't be in stable for a bit so ig this pr does make sense.

jmucc:
Well this workaround will still be needed. The code fix fixes the `JWT_SECRET` but the docs here concern the `PAGINATION_SECRET`.

